PR #21, opened 2017-05-02
Release/2.2.0

headione:


stephanlindauer:
did you check if the location exchange endpoint works for ios?

PR #24, opened 2017-10-10
Feature/less symbol opacity

headione:
hey @zutrinken , what do you think about this pull request. there was a request for less opacity of the bike symbols. this should improve the visibility of the swarm. android is also doing it.

i applied an opacity value of 0.7 of the current bike symbol. it looks good imho 🚲


headione:
an please ignore the travis fails... this isn't running at the moment ;)

zutrinken:
seems fine. I've updated the sketch file to get a preview of the opacity. Also recreated the assets from sketch to keep the design synced with the assets.

A different question would be if we should move from a bike symbol to a simple dot?

map

headione:
@zutrinken Thanks for your work so far! And yes, a simple dot looks way better IMHO. please commit these :) thanks 👍

stephanlindauer:
also, keep in mind: i already had quite a few requests to build a feature that lets you set your own marker color (maybe even tag with a name) and i think it's worth implementing. dots would be pretty simple to apply a certain color to i guess(?)

headione:
perfect, i'll make an release the next days. @stephanlindauer yep, coloring of icons shouldn't be a hard topic. maybe we need an svg later.

PR #25, opened 2018-11-25
Modern devices support

lennet:
This PR adds support for recent iOS devices like iPhone X, iPhone Xs etc.

I replaced the old Launch Screen Image with a storyboard and fixed some smaller layout issues that occurred on these newer devices


lennet:
The arrow button should be fixed now. I just reused the largest Icon which is available in 1x, 2x and 3x for the Launch screen 😅

headione:
Great! I'll merge! 🕺C

PR #26, opened 2019-01-15
Reimplemented Rules in Swift

lennet:
bildschirmfoto 2019-01-15 um 10 58 21
bildschirmfoto 2019-01-15 um 10 58 23

I started reimplementing all ViewControllers with Swift implementation based on the sketch file in the [design repo](https://github.com/criticalmaps/criticalmaps-design). This is the first PR for Rules.

The new design doesn't contain the Artwork anymore. Is that correct?


zutrinken:
Thanks for the PR! It looks great from a design perspective. I'd keep the artworks for the moment. Updating the content (text + new artworks) is a separate issue I think.

lennet:
@zutrinken Should I just add them above the description?

zutrinken:
@lennet Sure. Above should look good

PR #36, opened 2019-01-26
Reimplemented Chat in Swift

lennet:
This is a reimplementation of the Chat in Swift using the design specs from the design repo. 

I added a screen blocking loading indicator now because a loading indicator in the message cell and possible recovery states would have been too much work for now. But we should definitely consider it for the future.  

This PR also required new localized Strings. I used an online translator for the Italian translation so they might no be perfect.


CriticalMass/PLChatModel.m
```diff
@@ -1,130 +0,0 @@
...
-            co.isActive = YES;
-            co.timestamp = [NSString stringWithFormat:@"%lu", (long)[message objectForKey:@"timestamp"]];
-        }else{
-            // create chat object
-            PLChatObject *co = [[PLChatObject alloc] init];
-            co.identifier = key;
-            co.timestamp = [NSString stringWithFormat:@"%lu", (long)[message objectForKey:@"timestamp"]];
-            co.text = [[[message objectForKey:@"message"]
-                        stringByReplacingOccurrencesOfString:@"+" withString:@" "]
-                        stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
```
lennet: Are these character replacements still needed? @headione

headione: TBH, I'm not sure why I implemented this replacement - let's try it without :)

PR #42, opened 2019-02-04
New Navigation 

lennet:
To-Dos 
- [x] move the  segmented control to select between Chat & Twitter below the close button
- [x] improve selected visibility for navigation buttons 
- [x] add alt-text to navigation buttons
- [x] add a shadow to the navigation overlay 
- [x] move the maps legal button to a visible position


lennet:
@zutrinken Is the blur effect you use in the design specification the same as the default blur effect on iOS, or is it custom?

zutrinken:
> Is the blur effect you use in the design specification the same as the default blur effect on iOS, or is it custom?

Probably not. The style is 100% white with 80% alpha, background blur 4px, shadow Y 2px + blur 8px. It's supposed to look mainly white with a tiny glimpse of background shining through.

zutrinken:
@lennet @headione what's your opinion on the tabbar order? I could also imagine having it:

Location | Chat | Rules | Settings

headione:
Great work so far!

> @lennet @headione what's your opinion on the tabbar order? I could also imagine having it:
> 
> Location | Chat | Rules | Settings

I would also prefer this order tbh. Location should be first because it is the most important feature for this app imo. I also expect settings as the last button on the navbar. What do you think @lennet ?

lennet:
I totally agree location first and settings at the end makes sense!

@zutrinken Unfortunately the blur effect component on iOS is hard to customize and I don't want to create a custom blur but I'll try my best

lennet:
bildschirmfoto 2019-02-05 um 21 55 59
@zutrinken

zutrinken:
@lennet love it! I'd also switch chat and rules order if you agree.

PR #46, opened 2019-02-09
Watch App

lennet:
This is PR contains a proof of concept for a Critical Maps companion app for Watch OS.

All of the business logic can be shared between iOS and Watch OS. The Watch app is capable of accessing the GPS data and upload it to the server. It would even work as a standalone app without a connected iPhone on an LTE model. 

However, displaying annotations on a map is very limited with WatchKit 
- The map can only display up to 5 annotations at once but also doesn't support clustering out of the box
- The map doesn't support scrolling and zooming natively and a custom implementation is required

I use a combination of a grid-clustering to avoid overlaps of the annotations  and the k-means algorithm to cluster the currently visible elements into max 5 groups. I'm not an expert in Clustering so there is probably a better solution for that problem!

What do you think @zutrinken @headion ? Is that something which could make sense for us?

Possible next steps to complete the project

- [ ] add a way to disable GPS tracking on the watch
- [ ] clean up code & add tests for clustering 
- [ ] improve scrolling and zooming (Maybe just offering zoom and auto follow without manual scrolling might be a better experience)
- [ ] Improve UI
- [ ] add overlay if GPS is disabled
- [ ] the current grid clustering doesn't support coordinates around equator/ a translation to X,Y coordinates before clustering would be required


headione:
Wow, thanks for providing us this proof of concept! A watch app makes completely sense and critical maps is a good use case for that imo 👏

zutrinken:
I can get the app running on the simulator, but it doesn't show the map 😔 ... anyway, Watch support is a great idea!

Not sure what to do with the limited annotations. Maybe clustered is enough to see where the main mass is located. I can't really say.

More possible functions:
* Locate my position (icon in left bottom corner like apple maps)
* Sending a message via Dictation or Scribble

lennet:
@zutrinken I had the same issue. Opening the maps app on the iPhonne Simulator and restarting the watch simulator worked for me. https://forums.developer.apple.com/thread/69297

lennet:
What should we do with that PR. It surely needs more work but I don't want to abandon the PR.

How about merging it but excluding it from the Build so we are not shipping it to further work on it and test it? What do you think @zutrinken @mltbnz @headione ?

mltbnz:
I would be ok with that. Creates some technical depts for sure but who does not have them 😄 To keep track we can create an issue maybe

lennet:
@mltbnz @fbernutz 

There is currently no reason to keep this PR open. What do you think if we merge this PR into a new branch on the main repo so we can have it as a reference for future work and additionaly create a docs folder on master branch with a readme that explains the progress of this branch and what the problems with a watch app are?

mltbnz:
Think it is a good idea to create a feature branch for it and close this PR

mltbnz:
Gonna close the PR for now. I have pushed the branch to https://github.com/criticalmaps/criticalmaps-ios/tree/watch so that it is still around in case anyone wants to pick it up at a later point

Issue #31, opened 2019-01-22
Rule artworks are clipped


lennet:
Some Rule artworks are not displayed in the correct ratio 

Repro steps 

1. Navigate to Rules
2. Open "When Driving at the head .. "

Actual Result:
<img width="545" alt="bildschirmfoto 2019-01-22 um 13 09 24" src="https://user-images.githubusercontent.com/7677738/51534770-fcd6f280-1e46-11e9-9cfe-65b474aba687.png">

Expected Result:
![green](https://user-images.githubusercontent.com/7677738/51534893-59d2a880-1e47-11e9-8f94-783fcd5504c5.png)

PR #47, opened 2019-02-11
Fix issue #31 clipped images

huschu:
This PR fixes issue #31 

bildschirmfoto 2019-02-11 um 20 47 49
bildschirmfoto 2019-02-11 um 20 47 47


lennet:
Thank you for the PR! 🙏

Have you tried calculating that the offset (-30) from textContainerInset, linefragmentPadding and other properties instead of hardcoding the value?

huschu:
Thanks for the feedback! I have removed the hardcoded value.

PR #49, opened 2019-02-16
Active / Inactive Location Icon

### LINKED ISSUE #44: Active / Inactive Location Icon
When the map got moved and isn't following my own position the location icon should be outlined as in apple or google maps.

img_4265
img_4266

lennet:
fixes #44


lennet:
How about having a third state: follow + rotation ?

zutrinken:
@lennet there is a dedicated arrow icon for following. Instead of having the line above the tip I put it inside. You just have to rotate the icon so it's pointing north.

bildschirmfoto 2019-02-20 um 11 29 25

lennet:
@zutrinken I added the third mode and implemented a small scale transition. What do you think?

PR #51, opened 2019-03-04
Replaced PLAppDelegate with  a Swift implementation

lennet:
This PR removes PLAppDelegate and replaces it with a swift implementation. Additionally, I deleted the old prefix.pch and removed the swift code that was necessary for objc-interoperability

Not the most important PR. I just wanted to reach 100 swift :D


headione:
Awesome work! At this point it gets a little sad for me because all my legacy code is now removed :(

PR #60, opened 2019-03-30
Swift 5 Migration

### LINKED ISSUE #57: Swift 5 Migration

lennet:
I updated the Project to Swift 5 and fixed some project warnings 

fixes #57

### DIFF
--- a/CriticalMass/LocationManager.swift
+++ b/CriticalMass/LocationManager.swift
@@ -19,6 +19,9 @@ class LocationManager: NSObject, CLLocationManagerDelegate, LocationProvider {
             case .restricted,
                  .denied:
                 return .denied
+            @unknown default:
+                assertionFailure()
+                return .denied
             }
         } else {
             return .denied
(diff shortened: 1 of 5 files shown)


CriticalMass/LocationManager.swift
```diff
@@ -19,6 +19,9 @@ class LocationManager: NSObject, CLLocationManagerDelegate, LocationProvider {
             case .restricted,
                  .denied:
                 return .denied
+            @unknown default:
```
mltbnz: I think this is fine but I don't know if any new cases will be added to authorizationStatus. Or what is the reason you added it here?

lennet: Its not Markes as a frozen enumeration (not sure if they forgot it or wanna add cases in the future) so not handling the case gives an Warnung with swift 5 https://docs.swift.org/swift-book/ReferenceManual/Statements.html

mltbnz: Interesting 👍

PR #62, opened 2019-04-02
Add Version and Buildnumber to Settings

### LINKED ISSUE #55: Version and Build Number in Settings
Add the Artwork, version and build number to settings.

_Any other information to show here?_

mltbnz:
* adds a footer to the settings of the app with log, buildnumber and versionnumber
* adds acknowledgements and version (build)number to the system settings

closes #55


mltbnz:
Alright. I added you because of the discussion that was going on in the #55

headione:
Do you know why the build is failing @mltbnz @lennet ? 🤔

CriticalMass/SettingsViewController.swift
```diff
@@ -69,8 +69,34 @@ class SettingsViewController: UITableViewController {
...
+        super.viewDidLayoutSubviews()
+        guard let footerView = tableView.tableFooterView else {
+            return
+        }
+        let height = footerView.systemLayoutSizeFitting(UIView.layoutFittingCompressedSize).height
+        var footerFrame = footerView.frame
+        if height != footerFrame.size.height {
+            footerFrame.size.height = height
+            footerView.frame = footerFrame
+            tableView.tableFooterView = footerView
```
lennet: Why do we need to set the tableView.tableFooterView again? And ist the tableView.layoutIfNeeded() needed? I’m afraid that we could end up in endless layout cycles by triggering a layoutPass in viewDidLayoutSubviews().

CriticalMass/UIView+Nib.swift
```diff
@@ -0,0 +1,32 @@
...
+//  UIView+Nib.swift
+//  CriticalMaps
+//
+//  Created by Malte Bünz on 28.03.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import UIKit
+
+protocol NibInstantiatable {
```
lennet: 👏

CriticalMass/UIView+Nib.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+extension NibInstantiatable where Self: UIView {
+    
+    @discardableResult
+    static func fromNib() -> Self? {
+        let bundle = Bundle(for: self)
+        let nib = bundle.loadNibNamed(nibName(), owner: self, options: nil)
+        guard let view = nib?.first as? Self else {
```
lennet: Minor: we don’t need a guard here because the expected return type is an optional

Settings.bundle/com.mono0926.LicensePlist.plist
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lennet: Great Work! I love that we finally have third party license attribution!!

lennet:
> Do you know why the build is failing @mltbnz @lennet ? 🤔

Our tests seem to be unstable but it should be unrelated to this PR. I’ll take a look in the next days

CriticalMass/SettingsViewController.swift
mltbnz: It is only being set if the height actually has changed. So a new layout cycle is only triggered once. Better explained [here](https://useyourloaf.com/blog/variable-height-table-view-header/)

lennet: These lines are working for me or am I missing something? 
```swift
guard let footerView = tableView.tableFooterView else {
        return
}
let height = footerView.systemLayoutSizeFitting(UIView.layoutFittingCompressedSize).height
footerView.frame.size.height = height
```

CriticalMass/UIView+Nib.swift
mltbnz: `nib?.first` is of type `Any?` and a conversion to `Self` is not possible. So I have to use conditional cast

CriticalMass/SettingsViewController.swift
mltbnz: That doesn't produce a footer that is too big?

CriticalMass/UIView+Nib.swift
lennet:
```swift
return nib?.first as? Self
```
compiles for me

CriticalMass/SettingsViewController.swift
lennet: yes

CriticalMass/UIView+Nib.swift
mltbnz: It does here aswell 😀
Just thought it is more explicit. Will change to shorter version

CriticalMass/SettingsViewController.swift
mltbnz: Ok. I'll change it to the version you pointed out

mltbnz: Had to add the height check though so that we do not set the height every layout cycle

PR #64, opened 2019-04-04
Added Swiftformat to the project

lennet:
I thought it might helpful introduce a formatter/inter for a common styleguide. 
I've chosen Swiftformat without any custom configurations but I don't have strong opinions to other formatter/ linter and configurations as long as we have a consistent style guide.

What do you think @headione  @mltbnz?


mltbnz:
I think that is a very good idea

headione:
Swiftformat sounds good to me! 👍

mltbnz:
Not a big fan of the implicit `self` rule I guess but I can adopt to that. Has conflicts but apart from that fine

PR #70, opened 2019-04-10
Fixed decoding Android Messages

lennet:
The [Android app uses URL encoding](https://github.com/criticalmaps/criticalmaps-android/blob/d484b4b5266fdb81b7fd04e7c03adcf8878e8fc4/app/src/main/java/de/stephanlindauer/criticalmaps/model/ChatModel.java#L42)  for messages which currently doesn't get decoded properly. 

@stephanlindauer  is there a reason for the url encoding?


stephanlindauer:
@lennet 
can't remember why i did it this way :man_shrugging:

PR #82, opened 2019-04-20
Empty chat input fix

### LINKED ISSUE #73: Avoid empty messages
It's possible to send an empty message. Send button should be disabled until there's at least one character.

mltbnz:
disables send button when the chat message textField isEmpty.

fixes #73

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
 # Change Log
+
 Changelog for Critical Maps iOS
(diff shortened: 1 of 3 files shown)


mltbnz:
Merge the CHANGELOG was the biggest Challenge here. Every Editor adds the blank lines under ### lines of markdown files. I wanted to avoid them but now even github sneaked them in

PR #85, opened 2019-05-01
Added Network activity indicator support

lennet:
This will show a small spinner in the statusbar for active network requests

### DIFF
--- a/CriticalMass/NetworkActivityIndicatorHelper.swift
+++ b/CriticalMass/NetworkActivityIndicatorHelper.swift
@@ -1,0 +1,24 @@
+//
+//  NetworkActivityIndicatorHelper.swift
+//  CriticalMaps
+//
+//  Created by Leonard Thomas on 5/5/19.
+//
+
+import Foundation
+
+class NetworkActivityIndicatorHelper {
+    private var requestCounter = 0 {
+        didSet {
+            UIApplication.shared.isNetworkActivityIndicatorVisible = requestCounter > 0
+        }
+    }
+
+    public func didStartRequest() {
+        requestCounter += 1
+    }
+
+    public func didEndRequest() {
+        requestCounter -= 1
+    }
+}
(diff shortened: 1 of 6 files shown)


mltbnz:
This definetily helps the app. Though maybe it should live in the `NetworkOperator` since I think all components with network tasks are using that, right? Then also the Twitter and ChatController would benefit

lennet:
Oh, I totally forgot that we have instances that talk to the network layer directly. The Chat uses the request manager but Twitter doesn't.

CriticalMass/NetworkActivityIndicatorHelper.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by Leonard Thomas on 5/5/19.
+//
+
+import Foundation
+
+class NetworkActivityIndicatorHelper {
+    private var requestCounter = 0 {
+        didSet {
+            UIApplication.shared.isNetworkActivityIndicatorVisible = requestCounter > 0
```
mltbnz: Think this always needs to happen on the main queue

CriticalMass/NetworkActivityIndicatorHelper.swift
```diff
@@ -0,0 +1,24 @@
...
+//  NetworkActivityIndicatorHelper.swift
+//  CriticalMaps
+//
+//  Created by Leonard Thomas on 5/5/19.
+//
+
+import Foundation
+
+class NetworkActivityIndicatorHelper {
+    private var requestCounter = 0 {
```
mltbnz: maybe explicitly `UInt` type here since it should not be a negative value

lennet: Yeah I also thought about that but was to scared that this might crash at some point in production because it becomes -1 for some wired cases.. We could still do it or maybe at an assert so it only crashes for debug builds

mltbnz: 💡

PR #86, opened 2019-05-02
Improved message send 

lennet:
There was a possible bug that could prevent sending messages and show a loading indicator for ever if another network request is active. 

I also added a background task identifier for sending chat message to make sure that the messages get send after a user closes the app

### DIFF
--- a/CriticalMass/NetworkOperator.swift
+++ b/CriticalMass/NetworkOperator.swift
@@ -48,4 +48,8 @@ struct NetworkOperator: NetworkLayer {
         }
         task.resume()
     }
+
+    func cancelActiveRequestIfNeeded() {
+        session.invalidateAndCancel()
+    }
 }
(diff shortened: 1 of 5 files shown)


CriticalMass/NetworkOperator.swift
```diff
@@ -48,4 +48,8 @@ struct NetworkOperator: NetworkLayer {
         }
         task.resume()
     }
+
+    func cancelActiveRequestIfNeeded() {
+        session.invalidateAndCancel()
```
mltbnz: Works but maybe the `NetworkOperator` could return the current task and you just cancel that in the `RequestManager`. Though the Managers do not share a session, right?

lennet: We currently only have one Manager instance and it uses the same session for all requests. Changed the method name to requests to clarify that this method cancels all active requests

mltbnz: Though wouldn't canceling the session while in the chatController also cancel tasks that are being performed from the MapViewController?

lennet: Yes, so this method should only get called if you really want to cancel all tasks

PR #88, opened 2019-05-03
QR Code Pairing

lennet:
We are working on the feature on a feature branch (criticalmaps:friends) as it is a bigger feature.

This first PR contains a rough implementation for sharing and receiving links that enable Users to verify their friends based on a RSA signature. This allows us to continue dynamically change the Id and we don't have to store any metadata on the server.

#96 
qr


cbalster:
Hey! Great that you're tackling this feature!
We discussed a similar functionality a while back [here](https://github.com/criticalmaps/criticalmaps-android/issues/119).
From what I can gather from your PR, the ios-app is currently using the (constant) device ID to identify users. In the android version we changed that to a salted and hashed version that gets generated on app launch and is different every day (see [here](https://github.com/criticalmaps/criticalmaps-android/blob/master/app/src/main/java/de/stephanlindauer/criticalmaps/model/UserModel.java). We did this to not send out the unique device identifier and better protect you privacy if someone was to associate you with your id. I think this might be something we should also do on ios.
The idea discussed basically was to have a constant ID as soon as you use the feature. This would also solve the issue of removing the people following you: Just go back to changing ID's or generate a new one.
I think it would be a good idea to coordinate this between ios and android to a) release it to both apps at the same time and b) to make sure it's consistent and compatible between both versions.
I'd be happy with joining in the discussion here and as soon as it's specified enough, implement it in android.

lennet:
Thanks for the feedback! We should definitely add a changing Id to iOS as well. I’ll add my thoughts about the concept to the original issue in the Android repo :)

lennet:
@cbalster I updated my prototype and it can now handle changing IDs via RSA signatures. I used the name property to transmit the signature as a small workaround because id only allows 40 characters input. If we wanna move forward with this approach, we should probably add an optional signature  property to the API.

Unfollowing people is still possible by regenerating your local private key that is used to generate the signature.

CriticalMass/FriendsVerificationController.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  MapController.swift
```
mltbnz: maybe rename here although this is redundant info anyways. But we love XCode of course :D

mltbnz: I think that looks pretty good. Nice job

mltbnz: Can not really say a lot about implementation. I will dig into it though

lennet: Are you okay with merging it into the feature branch?

headione:
Looks pretty amazing @lennet . Within this implementation, is it possible to add multiple friends?

Could we please add it to Changelog? 😬

lennet:
Yes it is possible to have multiple friends. We‘ll continue working on this on a feature branch with multiple PRs and I‘ll update the Changelogs once its ready to be merged into master

Issue #89, opened 2019-05-05
Generate new deviceIDs after 24 hours 


lennet:
Currently, the deviceID which is Send to server is constant and only changes if a user reinstalls the app. We should change this I’d every 24 hours similar to what the android client does https://github.com/criticalmaps/criticalmaps-android/blob/master/app/src/main/java/de/stephanlindauer/criticalmaps/model/UserModel.java , to prevent that users can be tracked over multiple sessions and events

PR #91, opened 2019-05-05
Added changing device id depending on current date

lennet:
This PR adds a changing id based on the devices time similar to the android implementation https://github.com/criticalmaps/criticalmaps-android/blob/master/app/src/main/java/de/stephanlindauer/criticalmaps/model/UserModel.java

fixes. #89


lennet:
Sorry the PR isn't super clean its based on the work for #88, so it contains some changes that are not directly related to the current work item but I removed all the coded for the new feature

CriticalMass/IDStore.swift
```diff
@@ -0,0 +1,22 @@
...
+public class IDStore: IDProvider {
+    public init(currentDate: Date = Date()) {
+        let deviceID = UIDevice.current.identifierForVendor?.uuidString ?? UUID().uuidString
+
+        let format = DateFormatter()
+        format.dateFormat = "yyyy-MM-dd"
+        let dateString = format.string(from: currentDate)
+
+        id = (deviceID + dateString).md5
+    }
```
cbalster: Looks good to me. Note that in android we add 6 hours to the current time to shift the date to the next day. This way on the day of a CM (and the mass goes on until after 12pm) the ID is constant in case someone restarts the app after 12pm. tbh this is just a minor edgecase though.

lennet: How about events that start earlier? Wouldn’t it be better to check when the last session has started and if it’s more than 24 hours, renew the id without looking for the exact date

PR #93, opened 2019-05-09
Collect strings

mltbnz:
Intends to collect all NSLocalizedStrings in an extension to make them reusable and remove strings from logic controllers. Also moves and collects all URLs to the Constants container. 

* adds String+LocalizedStrings extension
* update strings in code with static lets from extension
* move URLs in Constants container
* change Constants type to enum so it is not instantiatable

🎨

### DIFF
--- a/CriticalMaps.xcodeproj/project.pbxproj
+++ b/CriticalMaps.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		0C2BD59F9813F04F7CDF34E6 /* libPods-CriticalMapsTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 2317F0770B48B2944D03CBE1 /* libPods-CriticalMapsTests.a */; };
 		8699EC9F4022B090262049D1 /* libPods-CriticalMaps.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 91EC6282437633054D179F01 /* libPods-CriticalMaps.a */; };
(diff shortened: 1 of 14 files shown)


lennet:
Nice! Maybe we should think about using code generation like https://github.com/SwiftGen/SwiftGen to generate String+NSLocalizedStrings

mltbnz:
Nice tool. Maybe we add https://github.com/JiriTrecak/Laurine to the app. For other use cases I don't know if we need to add a new dependecy to the project. But in general I don't know if adding another dependency is really needed for the ~ 30 Strings that are in the app.

mltbnz:
Btw. do you think we should add `static func` for these things?
```swift
var title: String {
        return NSLocalizedString("rules.title.\(rawValue)", comment: "")
    }

    var text: String {
        return NSLocalizedString("rules.text.\(rawValue)", comment: "")
    }
```

lennet:
> Nice tool. Maybe we add https://github.com/JiriTrecak/Laurine to the app. For other use cases I don't know if we need to add a new dependecy to the project. But in general I don't know if adding another dependency is really needed for the ~ 30 Strings that are in the app.

You are right, lets not introduce a dependency and re-evaluate this once maintaining this files becomes a lot of work

lennet:
> Btw. do you think we should add `static func` for these things?
> 
> ```swift
> var title: String {
>         return NSLocalizedString("rules.title.\(rawValue)", comment: "")
>     }
> 
>     var text: String {
>         return NSLocalizedString("rules.text.\(rawValue)", comment: "")
>     }
> ```

I'm not sure. I first thought no because the current code is compact and nice. 

However updating Rules by changing the content or the order is currently a lot of work and requires changes in all localization files. I think moving away from that logic and define the order in codes makes it easier to maintain

mltbnz:
So I just leave it like it is

PR #98, opened 2019-05-24
Chat input view fixes

lennet:
fixed:
- missing separator
- hidden by software keyboard on iPhone X devices
- missing placeholder after editing 

Bildschirmfoto 2019-05-24 um 14 50 49

### DIFF
--- a/CriticalMass/TextFieldWithInsets.swift
+++ b/CriticalMass/TextFieldWithInsets.swift
@@ -57,16 +57,6 @@ class TextFieldWithInsets: UITextField {
         return didBecomeFirstResponder
     }
 
-    // sets the backgroundColor when input ends
-    // kind of a hack sice the textField was always resetting its backgrundColor when it resigned firstResponder
-    override func resignFirstResponder() -> Bool {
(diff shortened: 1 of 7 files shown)


CriticalMass/TextFieldWithInsets.swift
```diff
@@ -57,16 +57,6 @@ class TextFieldWithInsets: UITextField {
         return didBecomeFirstResponder
     }
 
-    // sets the backgroundColor when input ends
-    // kind of a hack sice the textField was always resetting its backgrundColor when it resigned firstResponder
-    override func resignFirstResponder() -> Bool {
```
lennet: @mltbnz Is that save to remove? I don't see any issues without that hack and it fixes the missing placeholder after resigning first responder

mltbnz: You are right. Works also without it.

mltbnz: Then it is save :) Was there because of the weird way the TextField was acting with the theme change

PR #100, opened 2019-05-30
Dynamic type

mltbnz:
Adds DynamicType feature to the app.
The app will be more accessable for people with visual impairments when supporting DynamicType.

### DIFF
--- a/CriticalMass/UIFont+DynamicFont.swift
+++ b/CriticalMass/UIFont+DynamicFont.swift
@@ -1,0 +1,34 @@
+//
+//  UIFont+Scalable.swift
+//  CriticalMaps
+//
+//  Created by Malte Bünz on 30.05.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import UIKit
+
+extension UIFont {
+    /// Creates a System Font that is reacting to Font size changes from the OS.
+    ///
+    /// - Parameters:
+    ///   - fontSize: Initial font size.
+    ///   - weight: UIFont.Weight of the font.
+    /// - Returns: UIFont of Font SystemFont that is reacting to Dynamic Type changes.
+    static func scalableSystemFont(fontSize: CGFloat, weight: UIFont.Weight) -> UIFont {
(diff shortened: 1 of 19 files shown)


mltbnz:
#100 ✌️

lennet:
Great work! This is super important step to make the app more accessible :) 

We should try to use one of the system defined text styles if possible instead of redefining it https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/

We should also update the changelog

mltbnz:
Do you mean we should use the `UIFont.TextStyle`?

lennet:
Yes mapping sizes to Textstyles is the way to go. So instead of using for example instead of (fontSize: 17, weight: .regular) we should use the body text style. 

Only fonts that don’t match any existing text style should use the custom method

mltbnz:
I updates some of the font setting. Others use a custom style with for example `size: 15, weight: .bold`

CriticalMass/UIFont+DynamicFont.swift
```diff
@@ -0,0 +1,34 @@
...
+import UIKit
+
+extension UIFont {
+    /// Creates a System Font that is reacting to Font size changes from the OS.
+    ///
+    /// - Parameters:
+    ///   - fontSize: Initial font size.
+    ///   - weight: UIFont.Weight of the font.
+    /// - Returns: UIFont of Font SystemFont that is reacting to Dynamic Type changes.
+    static func scalableSystemFont(fontSize: CGFloat, weight: UIFont.Weight) -> UIFont {
```
lennet: It looks like we have 2 font styles that don't exist in the system font styles:
UIFont.scalableSystemFont(fontSize: 15, weight: .bold)
UIFont.scalableSystemFont(fontSize: 13, weight: .medium)

Maybe we can expose them in a central place and make scalableSystemFont private

mltbnz:
Wasn't shure about the gitHub cell but I missed the gpsDisabledOverlayView. Good catch. I think I will try to adjust the gitHub cell and then maybe @headione should take a look at it too :)

PR #101, opened 2019-06-01
Fixed missing update after failed request

lennet:
During yesterdays ride, my app was often stuck and didn't received any updates. Especially after coming back from background. 
I found a bug that we never set hasActiveRequests to false after a failed requests which could have caused that issue. The updated test case handles this case. 

I added some OSLogs for easier debugging in the future

### DIFF
--- a/CriticalMass/RequestManager.swift
+++ b/CriticalMass/RequestManager.swift
@@ -49,17 +49,15 @@ public class RequestManager {
     }
 
     private func defaultCompletion(for response: ApiResponse?) {
-        if let response = response {
-            DispatchQueue.main.async {
+        DispatchQueue.main.async {
+            defer {
                 self.hasActiveRequest = false
+            }
+            if let response = response {
                 self.dataStore.update(with: response)
 
                 Logger.log(.info, log: self.log, "Successfully finished API update")
-            }
-        } else {
-            DispatchQueue.main.async {
-                self.hasActiveRequest = false
-
+            } else {
                 Logger.log(.error, log: self.log, "API update failed")
             }
         }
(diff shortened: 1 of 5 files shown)


mltbnz:
Great! I also thought the app does not update riders positions after coming back from the background

lennet:
There might be more issues that might cause this issue. I'll collect some logs at tomorrows Sternfahrt

lennet:
I had no issues during the Sternfahrt :)

CriticalMass/RequestManager.swift
```diff
@@ -27,6 +28,8 @@ public class RequestManager {
     private var networkLayer: NetworkLayer
     private var idProvider: IDProvider
 
+    private var log = OSLog(subsystem: Bundle.main.bundleIdentifier!, category: "RequestManager")
```
mltbnz: Maybe the category can be more generic like „Networking“ and then outsource it in an extension of `OSLog` so it does not need to be an extra property of the RequestManager. But just a thought. Logger itself is nice. Just thought it is somehow strange to have the log in the class and then call have that for every class

mltbnz: meant someting like:

```swift
extension OSLog {
    private static var subsystem = Bundle.main.bundleIdentifier!

    static let network = OSLog(subsystem: subsystem, category: "Network")
}
```

and then:

```swift
Logger.log(.info, log: .network, "Successfully finished API update")
```

But it is just an opinion.
Merge at will

lennet: Good point, but let's do that once we need to add more logging in the future and adding the abstraction is needed

PR #102, opened 2019-06-13
DarkModelOverlayFix

mltbnz:
Adds mapOverlays aboveRoads instead of aboveLabels so that
Screenshot 2019-06-13 at 09 26 42
becomes 
Screenshot 2019-06-13 at 09 28 04


mltbnz:
I think it might when accessing the overlays above the street I thinkt you can get the labels, remove them and add your own. But without trying it sounds costly. Maybe we give it a try without first

mltbnz:
I think it might when accessing the overlays above the street I thinkt you can get the labels, remove them and add your own. But without trying it sounds costly. Maybe we give it a try without first

Issue #41, opened 2019-02-03
Observing Mode


lennet:
Observing mode is for users that want to look at the map without participating in the critical mass. 

<img src="https://user-images.githubusercontent.com/888679/55790007-c579f880-5abb-11e9-9f2a-0e7796a2b29a.png" alt="screenshot" width="375" height="283" />

PR #104, opened 2019-06-15
Observation mode 

lennet:
This PR adds the Observation mode (#41) and replaces the old GPS switch to disable posting updates to the backend. I also updated the UI for the No GPS overlay which no has a button that links directly to the system settings 

Light mode:

Bildschirmfoto 2019-06-15 um 23 20 58Bildschirmfoto 2019-06-15 um 23 21 00

Dark mode:

Bildschirmfoto 2019-06-15 um 23 21 10Bildschirmfoto 2019-06-15 um 23 21 06

### DIFF
--- a/CriticalMass/BlurryOverlayView.swift
+++ b/CriticalMass/BlurryOverlayView.swift
@@ -8,8 +8,46 @@
 
 import UIKit
 
+@IBDesignable
+class RoundedButton: UIButton {
+    override func awakeFromNib() {
+        configureUI()
+    }
+
+    override func prepareForInterfaceBuilder() {
+        configureUI()
+    }
+
+    private func configureUI() {
+        setTitleColor(.black, for: .normal)
+        backgroundColor = .yellow100
+        layer.cornerRadius = 24
+        titleLabel?.font = UIFont.scalableSystemFont(fontSize: 17, weight: .bold)
+        titleLabel?.adjustsFontForContentSizeCategory = true
+    }
+}
+
 class BlurryOverlayView: UIView, IBConstructable {
+    @objc
+    dynamic var gradientBeginColor: UIColor = .black {
+        didSet {
+            updateGradient()
+        }
+    }
+
+    @objc
+    dynamic var gradientEndColor: UIColor = .white {
+        didSet {
+            updateGradient()
+        }
+    }
+
     @IBOutlet private var messageLabel: UILabel!
+    @IBOutlet var titlelabel: UILabel!
+
+    override class var layerClass: AnyClass {
+        return CAGradientLayer.self
+    }
 
     public var message: String? {
         didSet {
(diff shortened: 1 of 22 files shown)


CriticalMass/Preferences.swift
```diff
@@ -7,17 +7,33 @@
...
+    init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    func save(_ isEnabled: Bool) {
+        defaults.set(isEnabled, forKey: defaultsKey)
+        NotificationCenter.default.post(name: Notification.observationModeChanged, object: isEnabled)
+    }
+
+    func load() -> Bool? {
```
mltbnz: Should `save` and `load` maybe be private? Otherwise I think naming could be improved

CriticalMass/SettingsSwitchTableViewCell.swift
```diff
@@ -7,18 +7,16 @@
 
 import UIKit
 
-protocol SettingsSwitchCellConfigurable {
-    func configure(isOn: Bool, handler: SettingsSwitchHandler?)
+protocol Switchable {
```
mltbnz: Is `Switchable` to keep the SettingsController more adjustable and maybe add other switch cell later?

CriticalMass/SettingsViewController.swift
```diff
@@ -102,48 +91,31 @@ class SettingsViewController: UITableViewController {
         let acttion = section.models[indexPath.row].action
```
mltbnz: Typo `acttion`

CriticalMass/SettingsViewController.swift
```diff
@@ -52,28 +52,18 @@ class SettingsViewController: UITableViewController {
         }
     }
 
-    // MARK: Actions
-
-    @IBAction func gpsCellAction(_: UISwitch) {}
```
mltbnz: I was actually a fan of having the `@IBAction` in the controller. Makes it easy to understand. Where is the switch target now?

CriticalMass/BlurryOverlayView.swift
```diff
@@ -8,15 +8,74 @@
...
+        updateGradient()
+    }
+
+    public func set(title: String, message: String) {
+        titlelabel.text = title
+        messageLabel.text = message
+    }
+
+    @IBAction private func didTapSettingsButton(_: Any) {
+        UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!)
```
mltbnz: Hmm. Maybe we should add a controller for the BlurryOverlay, since it now does logic. wdyt?

mltbnz: Or we use the responderchain and set the action on the undelying controller?

CriticalMass/SettingsSwitchTableViewCell.swift
lennet: Yes, my main though replace the repetitive code in SettingsViewController with a common interface

CriticalMass/SettingsViewController.swift
lennet: The cell no gets a switchable element that can handle the update

CriticalMass/BlurryOverlayView.swift
lennet: its no a lot of logic but you are right. I'll let the controller handle the action

lennet:
@zutrinken 

Bildschirmfoto 2019-06-16 um 20 38 15

CriticalMass/SettingsSwitchTableViewCell.swift
mltbnz: Ok 👌

CriticalMass/BlurryOverlayView.swift
mltbnz: True. Very little logic

PR #107, opened 2019-06-23
Added user setting to ignore the iOS sleep timer 

lennet:
In this PR I added a Switch to Settings that will disable the iOS sleep timer while the app is open

C1CBC314-6046-46A1-976A-051849702032

If that’s something we wanna have in the app I still need to work on these tasks:
- fix the height for settings cells without subtitles 
- update change log 
- add translations 
- maybe adding another level of abstraction for switchable and userdefaults to minimize duplicated code with the observer more logic


mltbnz:
Is that a feature the users want? Or was it requested? 
I really can not say if it is usefull or not for CM participants.

From a tech perspective it drains a lot of power apple advisess against it.
From a riders perspective it might be cool to have a look at the map at all times. I wouldn't use it since these modes kind of distract me from the road.

cbalster:
FYI: We had this as a default in the android app for ages. As soon as I
made it an option with default=off someone (who didn't see it in the
options) gave us a 1 star review, complaining about it :D

mltbnz:
Interesting. I guess because of battery usage?

cbalster:
Ah, sorry. No, they complained because they thought the feature to keep the screen on was removed!  People seem to want that as an option.

lennet:
I don’t know of any iOS users that requested that feature but I would like to have it myself :D 

I would definitely offer it as an optional feature which is disabled by default but it might be nice for some users that want to see the map or are waiting for a chat response. 

I‘m using a mount for my phone during CM and with that unlocking while cycling is pretty annoying because Face ID fails and I have to enter my pin code manually. I would personally prefer that the App only sleeps if active the devices sleep button.

mltbnz:
> I don’t know of any iOS users that requested that feature but I would like to have it myself :D
> 
> I would definitely offer it as an optional feature which is disabled by default but it might be nice for some users that want to see the map or are waiting for a chat response.
> 
> I‘m using a mount for my phone during CM and with that unlocking while cycling is pretty annoying because Face ID fails and I have to enter my pin code manually. I would personally prefer that the App only sleeps if active the devices sleep button.

I can see it totally makes sense for that case.
The arguments that came up already convinced me 👍

mltbnz:
Maybe the switch then should also have an explanation

lennet:
I'm closing the PR for now but I'll keep the branch in case we wanna have that feature in the future

PR #108, opened 2019-06-24
Social selection fix

### LINKED ISSUE #53: SegmentedControl in SocialViewController slipping under Toolbar in iOS <11.0
Looks like the segmentedControl hides under the toolbar for example on iOS 10.3.1
Screenshot 2019-03-12 at 16 06 26

mltbnz:

This PR will fix #53 

Before:
Screenshot 2019-06-24 at 08 38 05

After:
Screenshot 2019-06-24 at 08 39 06

### Before merging the PR


CriticalMass/AppController.swift
```diff
@@ -60,7 +60,7 @@ class AppController {
     }
 
     private func getSocialViewController() -> SocialViewController {
-        return SocialViewController(chatViewController: getChatViewController, twitterViewController: getTwitterViewController)
+        return SocialViewController(chatViewController: getChatViewController(), twitterViewController: getTwitterViewController())
```
lennet: What are the benefits here from removing the lazy loading?

mltbnz: I don't see why they should be lazy. Actually I just changed the property inside the SocialController to be a `let` rarther then to be a closure so it makes code more readable in my eyes. What is the benefit here of having them lazy? It does not need to capture the AppController instance, right? Would opt-in for constants there anyhow I think

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changelog for Critical Maps iOS
 
 ### Fixed
 
+- SocialSegment sliding under NavigationBar bug in iOS 10
```
lennet: This should be added to Unreleased

CriticalMass/AppController.swift
lennet: Mostly smaller memory optimizations. By loading the content lazily only the current content is in memory and on switching the tab the old content get's released

CHANGELOG.md
mltbnz: Ups. Thought I did

CriticalMass/AppController.swift
mltbnz: Hmm. That is true. Might also happen the user never opens the socialController. So that could be saved. I do think though that the pattern is mostly used for computational heavy allocations and if some properties are only there after initialization of the object that holds the lazy var. Did you ever profile the memory?

CHANGELOG.md
mltbnz: Think it is:
https://github.com/mltbnz/criticalmaps-ios/blob/socialSelectionFix/CHANGELOG.md

PR #110, opened 2019-06-24
Fixed app for assistive technology users

lennet:
We were using MKMapView as the root view of our app which has isAccessibilityElement set to true by default. A view that is an accessibility element can't have any accessible child elements which made the hall app (except for the map) inaccessible for VoiceOver and Switch Control users. 

Using a normal UIView as a root view and adding the Map as a Subview solves that issue

### Before merging the PR

~~- [ ] If applicable, does the PR contain enough unit / UI tests?~~
~~- [ ] If applicable, did you add a before / after screenshot of the feature?~~


CriticalMass/MapViewController.swift
```diff
@@ -90,6 +86,10 @@ class MapViewController: UIViewController {
     }
 
     private func configureMapView() {
+        mapView.frame = view.frame
+        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```
mltbnz: I would opt-in for NSLayoutConstraints here

PR #111, opened 2019-06-25
Update of French localizable.strings

AlbanSagouis:
Translation based on the English version
Added missing strings
Deleted outdated strings
Improved a few translations
Changed string order and format to match row numbers with the English version


lennet:
Thanks!

PR #119, opened 2019-08-16
Result all the things

mltbnz:
This PR will introduce Swifts Result type to the network layer

### Before merging the PR


CriticalMass/PostLocationRequest.swift
```diff
@@ -0,0 +1,17 @@
...
+//  PostLocationRequest.swift
+//  CriticalMaps
+//
+//  Created by Malte Bünz on 16.08.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import Foundation
+
+struct PostLocationRequest: APIRequestDefining {
```
lennet: similar to the comment about GetLocationsAndChatMessagesRequest

CriticalMass/ErrorHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Created by Malte Bünz on 16.08.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import Foundation
+
+struct ErrorHandler {
+    static let `default` = ErrorHandler()
+
+    let genericErrorMessage = "Sorry, something went wrong"
```
lennet: this need to be localised

CriticalMass/ErrorHandler.swift
```diff
@@ -0,0 +1,27 @@
...
+    func handleError(_ error: Error?) {
+        if let errorDescription = error?.localizedDescription {
+            presentToUser(message: errorDescription)
+        } else {
+            presentToUser(message: genericErrorMessage)
+        }
+    }
+
+    private func presentToUser(message: String) {
+        print(message)
```
lennet: Do we want to add alerts in this case? Although we shouldn't present alerts if the app is running in the background

CriticalMass/NetworkOperator.swift
```diff
@@ -18,36 +18,55 @@ struct NetworkOperator: NetworkLayer {
...
-        var request = URLRequest(url: url)
-        request.httpMethod = "POST"
-        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
-        request.httpBody = bodyData
-        dataTask(with: request, decodable: decodable, customDateFormatter: nil, completion: completion)
+    func post<T: APIRequestDefining>(request: T, bodyData: Data, completion: @escaping ResultCallback<T.ResponseDataType>) {
+        var urlRequest = request.makeRequest()
+        urlRequest.httpBody = bodyData
+        dataTask(with: urlRequest) { result in
+            switch result {
```
lennet: minor: there is some repetition in the result handling between post and get. I'm wondering if we could simplify that

CriticalMaps.xcodeproj/project.pbxproj
```diff
@@ -1028,7 +1092,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
-				DEVELOPMENT_TEAM = 5YLLXUZBZ2;
+				DEVELOPMENT_TEAM = BP8836JVQR;
```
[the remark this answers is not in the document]
mltbnz: Right. I missed that

CriticalMass/GetLocationsAndChatMessagesRequest.swift
```diff
@@ -0,0 +1,17 @@
...
+//  GetLocationsAndChatMessagesRequest.swift
+//  CriticalMaps
+//
+//  Created by Malte Bünz on 11.08.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import Foundation
+
+struct GetLocationsAndChatMessagesRequest: APIRequestDefining {
```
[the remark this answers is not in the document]
mltbnz: I did not find a solution with the protocol to add a default here. Do you know a way?

CriticalMass/NetworkOperator.swift
mltbnz: Good point. I was thinking of something like passing in the body as an optional and then only adding it if it has a value

CriticalMass/GetLocationsAndChatMessagesRequest.swift
```diff
@@ -0,0 +1,17 @@
...
+    typealias ResponseDataType = ApiResponse
+    var baseUrl: URL
+    var paths: [String]
+    var headers: HTTPHeaders?
```
lennet:
```suggestion
    var headers: HTTPHeaders? = nil
```

CriticalMass/GetLocationsAndChatMessagesRequest.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Malte Bünz on 11.08.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import Foundation
+
+struct GetLocationsAndChatMessagesRequest: APIRequestDefining {
+    typealias ResponseDataType = ApiResponse
+    var baseUrl: URL
+    var paths: [String]
```
lennet:
```suggestion
    var paths: [String] = []
```

CriticalMass/GetLocationsAndChatMessagesRequest.swift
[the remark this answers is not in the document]
lennet: added two example suggestions

CriticalMass/ErrorHandler.swift
mltbnz: I want to add something visual after this PR is merged. For now I am not sure how it looks like. Maybe also @zutrinken has an idea of error states

CriticalMass/NetworkOperator.swift
mltbnz: Oh. I confused your reply. I did not find a good (quick) solution. I want to find something tho and will park the ticket some time until I can work on a solution

CriticalMass/ErrorHandler.swift
mltbnz: I want to add something visual after this PR is merged. For now I am not sure how it looks like. Maybe also @zutrinken has an idea of error states

CriticalMass/GetLocationsAndChatMessagesRequest.swift
mltbnz: I added a struct `Endpoint` that solves the issue I think. I also updated the creation logic of the request

Issue #113, opened 2019-06-28
iOS 13 Support


lennet:
## Modal presentations

iOS 13 has a new default modal transition.

![modal](https://user-images.githubusercontent.com/7677738/60342407-ab143480-99b1-11e9-8f87-f0b9af553b01.gif) 

Do we need some design changes here @zutrinken . Using the old full screen presentation would also be possible

## Colors - Darkmode 

Our Theming/ Darkmode implementation breaks for some components. Maybe we should adopt to the system Darkmode for iOS 13 Users.
<img width="456" alt="Bildschirmfoto 2019-06-28 um 14 33 36" src="https://user-images.githubusercontent.com/7677738/60342432-b9fae700-99b1-11e9-9794-ac12d502a12d.png">
<img width="456" alt="Bildschirmfoto 2019-06-28 um 14 33 25" src="https://user-images.githubusercontent.com/7677738/60342450-c121f500-99b1-11e9-9a36-54ef424d0104.png">

## Swift Package Manager

It would be nice in my opinion to replace our current dependency manager cocoapods with the Swift Package Manager that is built into Xcode 11. That would allow us to build the project without needing any additional tooling

zutrinken:
New modal looks good. One detail is the background color behind the map. Would make it black to see some depth.

mltbnz:
I would love to adopt the dark mode to native support

lennet:
<img width="386" alt="Bildschirmfoto 2019-08-10 um 09 26 16" src="https://user-images.githubusercontent.com/7677738/62819117-f5600800-bb50-11e9-8aed-73f12aee72d8.png">
 
Thats how the chat / twitter control looks like with the current beta @zutrinken

mltbnz:
I noticed that the navigationBar in the socialcontroller looks the same on `develop`. In the designs it is never translucent but we set it to be in the light mode. Was that by mistake?

PR #123, opened 2019-09-12
Introduced Swift Packager Manager as a dependency manager

lennet:
I started replacing Cocoapods with Swift Packager Manger as it's build right into Xcode 11

Open Questions:
- [x] Swift format
- [x] License Plist generation

#113


lennet:
Yes both work without cocoapods but I still need to do that 😅

lennet:
bundled swift-format into the repo and removed all remaining cocoapods files. 
license-plist doesn't support SwiftPM [yet](https://github.com/mono0926/LicensePlist/issues/107)

mltbnz:
LicensePlist should work now 🥳
https://github.com/mono0926/LicensePlist/releases/tag/2.6.0

mltbnz:
Tried it yesterday and it seems like LicensePlist still has some issues...

Also to get rid of the `Pod` folder we should execute 
```shell
pod deintegrate
``` 

and also delete the `.xcworkspace`

lennet:
I did that. The problem ist that many cocoapods files are part of our .gitignore file so they don't get automatically deleted. I removed them from our .gitignore to make the more visible. After deleting the files LicensePlist works for me

mltbnz:
Works. Nice.
But now I have this https://github.com/mono0926/LicensePlist/issues/105

lennet:
It seems to be unrelated to this PR as the bug is already in master. I think create an issue and a follow up PR once we have a fix

mltbnz:
Ok. Then merge at will

PR #124, opened 2019-09-21
Fix NavigationBar colors under iOS 13

mltbnz:
This should fix navigationbar colors under iOS 13 which currently don't look right in XCode Sims.
Let's hope 13.1 does not change that :D

**Screenshots**
Screenshot 2019-09-21 at 15 43 50Screenshot 2019-09-21 at 16 06 41

Screenshot 2019-09-21 at 15 44 03Screenshot 2019-09-21 at 15 31 51

### Before merging the PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-language: objective-c 
-osx_image: xcode10.3
+language: objective-c
+osx_image: xcode11
 
 before_install:
   - bundle update
 
 script:
-    - bundle exec fastlane test
+  - bundle exec fastlane test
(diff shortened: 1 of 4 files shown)


mltbnz:
True. I missed those smaller titles.
I like the new style by I can check what it takes to rebuild the old one

zutrinken:
@mltbnz Thanks for checking ... I agree, the new style looks pretty good. But it doesn't match our UI. Would need a bunch of changes on our side (lighter/rounder icons, softer colours, ...). Therefore I'd rather stick with the old/current style.

PR #125, opened 2019-09-21
Adopt dark mode

mltbnz:
This PR will set the Apps Theme to the system Theme under iOS13, when the user did not set a theme before. Until the user sets a theme in the app, the systemwide userStyle will be used.

Screenshot 2019-09-21 at 18 47 24
Screenshot 2019-09-21 at 18 45 47
Screenshot 2019-09-21 at 18 45 54

### Before merging the PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
-language: objective-c 
-osx_image: xcode10.3
+language: objective-c
+osx_image: xcode11
 
 before_install:
   - bundle update
 
 script:
-    - bundle exec fastlane test
+  - bundle exec fastlane test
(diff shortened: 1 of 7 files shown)


mltbnz:
🙀 I see. 
I like the option to have only some apps in dark mode. Would want to support it for older versions as well.

lennet:
I agree, we should make dark mode an iOS 13 + feature. It makes is easier to maintain and we could adapt the system api and remove our hacks

mltbnz:
Would remove some of the (seeing it now) messy code for sure. If we do we have it only for iOS 13+ users we should wait until maybe Jan. I think.

@zutrinken The issue in the navigationBar is fixed if we merge #124 first

PR #135, opened 2019-10-03
Fix table view header footer background color warning

### LINKED ISSUE #76: SettingsFooterView uses deprecated API
"Setting the background color on UITableViewHeaderFooterView has been deprecated. Please set a custom UIView with your desired background color to the backgroundView property instead."

maxxx777:
This PR fixes #76. 
After overriding `backgroundColor` var in `UITableViewHeaderFooterView`, the message `Setting the background color on UITableViewHeaderFooterView has been deprecated. Please set a custom UIView with your desired background color to the backgroundView property instead.` doesn't appear in the logs anymore.

### Screenshots (before the changes)
IMG_0006
IMG_0005

### Screenshots (after the changes)

IMG_0011
IMG_0012

### Before merging the PR


lennet:
Thank you for the contributing!

That's indeed removing the Error message. However it isn't solving the issue that we don't have a backgroundColor as the backgroundView property is nil 

Bildschirmfoto 2019-10-03 um 17 28 17

Setting a plain UIView as backgroundView might solve that

maxxx777:
> Thank you for the contributing!
> 
> That's indeed removing the Error message. However it isn't solving the issue that we don't have a backgroundColor as the backgroundView property is nil
> 
> Bildschirmfoto 2019-10-03 um 17 28 17
> 
> Setting a plain UIView as backgroundView might solve that

good point 👍
did the update.

Issue #136, opened 2019-10-03
Add Snapshot testing


maxxx777:
Add Snapshot testing library https://github.com/pointfreeco/swift-snapshot-testing (it supports installation with SPM).

Pros:
- automatically test how all the screens/views we need to test look on different devices.
- automatically test how all the screens/views we need to test look in both light and dark modes.
- automatically test how all the strings on the screens/views we need to test look
- automatically test how all the strings on the screens/views we need to test look in different localizations (if we have different localizations)
- don't need to submit before/after screenshots with every PR, because it will be tested automatically

maxxx777:
@lennet @headione @mltbnz what do you think? I could implement it.

lennet:
Thanks for the suggestion! I really like the Idea!  

Not sure if it works with our current theming solution as we are currently using UIAppearance and I'm not sure if applying a Theme also works without being part of a presented UIWindow.

I would love to see how it performs, how stable the results are and how long it takes to generate the reference images. Maybe we could try it first with one sample view like the TweetTableViewCell.

maxxx777:
thanks, I will create a branch with PoC and then we can see the results.

mltbnz:
Should we close this since there are snapshot tests now?! Thanks again for adding those 🙏

PR #139, opened 2019-10-05
Add snapshot testing

maxxx777:
This PR provides PoC for proposal #136.
- add a new target for Snapshot tests
- add snapshot-testing swift package
- add snapshot tests for TweetTableViewCell as an example

### Before merging the PR


README.md
```diff
@@ -53,7 +53,9 @@ In general, we follow the "fork-and-pull" Git workflow.
 4.  **Push** your work back up to your fork
 5.  Submit a **Pull request** so that we can review your changes
 
-NOTE: Be sure to merge the latest from "upstream" before making a pull request!
+NOTES: 
+- Be sure to merge the latest from "upstream" before making a pull request!
+- Please use "iPhone 5s (iOS 12.2)" simulator for snapshot testing (because Travis is running the tests with this device)
```
lennet: I'm wondering if we should specify the simulator in our Fastfile instead of using the travis default simulator which might change at some point?

CriticalMapsSnapshotTests/MockThemeController.swift
```diff
@@ -0,0 +1,33 @@
...
+class MockThemeController: ThemeController {
+    static let shared = MockThemeController()
+    
+    private init() {
+        let store = MockThemeStore()
+        super.init(store: store)
+    }
+}
+
+//TODO: move MockThemeStore class to shared file for both CriticalMassTests and CriticalMapsSnapshotTests targets
```
lennet: Are you planing to resolve that TODO before we merge it to master. We should create an issue if we need to follow up

CriticalMass/FormatDisplay.swift
```diff
@@ -29,8 +29,11 @@ private extension DateComponents {
 }
 
 enum FormatDisplay {
+    // inject current date to be able to keep the same time difference between tweet date and current date
+    static var currentDate = Date()
```
lennet: Not a fan of having a static var here and would prefer injecting the date as an argument or initialiser but nothing which we can easily do with the current architecture here but that should be okay for no

CriticalMapsSnapshotTests/TweetTableViewCellSnapshotTests.swift
```diff
@@ -0,0 +1,40 @@
...
+    private let tweetDate = Date(timeIntervalSince1970: 1530230956)
+    private let CellReuseIdentifier = "Cell"
+    
+    override func setUp() {
+        tableViewController.tableView.register(TweetTableViewCell.nib, forCellReuseIdentifier: CellReuseIdentifier)
+        // use constant date in the past
+        FormatDisplay.currentDate = Date(timeIntervalSince1970: 1530240956)
+    }
+
+    func testTweetTableViewCell() {
```
lennet: Love how simple the test case looks like!

CriticalMapsSnapshotTests/TweetTableViewCellSnapshotTests.swift
```diff
@@ -0,0 +1,40 @@
...
+        // given
+        let tweet = Tweet(text: "Hello World",
+                          created_at: tweetDate,
+                          user: TwitterUser(name: "Bar",
+                                            screen_name: "Foo",
+                                            profile_image_url_https: "haa"))
+        let cell = tableViewController.tableView.dequeueReusableCell(withIdentifier: CellReuseIdentifier) as! TweetTableViewCell
```
lennet: Do we need a UITableView here?  Can't we init the cell directly?

README.md
maxxx777: Good point! Specified the device in Fastfile and updated README.

CriticalMass/FormatDisplay.swift
maxxx777: Yep, maybe we could refactor it in the future.

CriticalMapsSnapshotTests/MockThemeController.swift
maxxx777: Resolved by sharing some mock classes between test targets

CriticalMapsSnapshotTests/TweetTableViewCellSnapshotTests.swift
maxxx777: Very good suggestion!

maxxx777:
@lennet thanks for the comments. did the updates, please take a look again

CriticalMapsMocksForTests/MockThemeController.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  SnapshotTester.swift
+//  CriticalMapsSnapshotTests
+//
+//  Created by MAXIM TSVETKOV on 05.10.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+@testable import CriticalMaps
+import SnapshotTesting
```
lennet: Do we need to import SnapshotTesing? It is currently unavailable for the unit test target

lennet:
Sorry for the delayed response. I fixed the merge conflict but it seems we have to update the device we use because Travis doesn’t has the iPhone 5s installed

maxxx777:
> Sorry for the delayed response. I fixed the merge conflict but it seems we have to update the device we use because Travis doesn’t has the iPhone 5s installed

No worries, thanks for merging the conflicts. I updated the device to iPhone 8.

PR #140, opened 2019-10-06
Add UITest Infrastructure

### LINKED ISSUE #84: UI Tests
It would be nice to have a infrastructure for UI tests. We could use that for testing but also screenshot to have a quick overview how the app looks like in different languages and device types

fbernutz:
Fixes #84. 

This includes the following changes: 
- Add new Target for UITests
- Add UITest which navigates through all screens and tests the navigation, if all screens are reachable
- Add fastlane action to run `bundle exec fastlane screenshots` to generate screenshots from every screen in all available languages and different devices
- Check in the configurated schemes to run UI Tests separately from Unit Tests 
   - CriticalMaps
   - CriticalMapsTests
   - CriticalMapsUITests
- Update Fastlane

### Before merging the PR


CriticalMapsUITests/CriticalMapsUITests.swift
```diff
@@ -0,0 +1,62 @@
...
+//  CriticalMapsUITests.swift
+//  CriticalMapsUITests
+//
+//  Created by Felizia Bernutz on 06.10.19.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import XCTest
+
+class CriticalMapsUITests: XCTestCase {
```
lennet: I'm wondering if we should run that test for one language and one device on the CI. 

We had a bug in the past were the app wasn't operable anymore with VoiceOver and Switch control and this test would have caught that.

CriticalMapsUITests/SnapshotHelper.swift
```diff
@@ -0,0 +1,303 @@
...
+//
+//  Created by Felix Krause on 10/8/15.
+//
+
+// -----------------------------------------------------
+// IMPORTANT: When modifying this file, make sure to
+//            increment the version number at the very
+//            bottom of the file to notify users about
+//            the new SnapshotHelper.swift
+// -----------------------------------------------------
```
lennet: Is there a Swift package or any other keep the connection to automatically get updates in the future?

CriticalMass/AppDelegate.swift
```diff
@@ -25,10 +25,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+    private func handleLaunchArguments() {
+        #if DEBUG
+        let arguments = ProcessInfo().arguments
+
+        if arguments.contains("SKIP_ANIMATIONS") {
+            UIView.setAnimationsEnabled(false)
+        }
+        #endif
+    }
```
lennet: There some more logic for XCTest in the `func application(_: UIApplication, didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?) -> Bool ` 

Maybe we should move the test related setup code to one place even though it's not 100% related

CriticalMapsUITests/CriticalMapsUITests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    func testScreenshots() {
+        let app = XCUIApplication()
+        snapshot("01_map")
+
+        let closeButton = app.navigationBars.buttons.element(boundBy: 0).firstMatch
+        let backButton = closeButton
+
+        XCTContext.runActivity(named: "Show Chat") { _ in
```
lennet: 🤯 Didn't know about that API

fbernutz: I learned about it some months ago and it makes the test report in Xcode so much more structured. Even the code! I always used comments to mark different use cases, but this is great for readability as well 👍

CriticalMapsUITests/SnapshotHelper.swift
fbernutz: Mhm... I don't think so. I didn't find anything about it online. But when one UITest runs in the CI, the CI would fail if someone has updated fastlane but not the SnapshotHelper. 

Because I forgot to update the SnapshotHelper every time in one of my projects, I added a script `update-dependencies.sh` or whatever name you like, where I collected all commands you have to run to update the dependencies. So you can run `bundle update` and `bundle exec fastlane snapshot update` in there. But yeah, this isn't really nice as well..

CriticalMapsUITests/CriticalMapsUITests.swift
fbernutz: Yeah, sounds like a good idea!  I'll add another fastlane action which can run on the CI 👍

fbernutz: fyi: There are now two jobs running in parallel. 

image

CriticalMapsUITests/SnapshotHelper.swift
lennet: Sounds like a good tradeoff. 

Created an issue for that #142

CriticalMass/AppDelegate.swift
fbernutz: I tried my best in finding a good name to combine both. If you have another idea for the naming, feel free to make a suggestion 😄

fbernutz:
I implemented your feedback, please have a look again  🙂 
Sadly, the pipeline failed again for the Unit Tests, I have no clue what the issue is. Looks like a retry could "help". Maybe you have an idea?

`Operation not permitted - /Users/travis/Library/Application Support/CallHistoryTransactions`

```xcodebuild: error: Could not resolve package dependencies:
  Error Domain=NSCocoaErrorDomain Code=513 "“SDWebImage” couldn’t be removed because you don’t have permission to access it." UserInfo={NSFilePath=/Users/travis/Library/Developer/Xcode/DerivedData/CriticalMaps-akluakkxjuxftqeuxfxonhxinpjk/SourcePackages/checkouts/SDWebImage, NSUserStringVariant=(
    Remove
), NSUnderlyingError=0x7fd25960f640 {Error Domain=NSPOSIXErrorDomain Code=66 "Directory not empty"}}```

fbernutz:
I don't know whats wrong with the CI. Running `bundle exec fastlane test` locally and running the Unit Tests in Xcode both succeed for me.

lennet:
Hmm the Unit test pipeline shouldn't fail but it seems to work now. We should keep an eye on that for the next PR but the UITest seem to work fine 🥳

lennet:
🚢 Lets merge it

mltbnz:
🚀

Issue #142, opened 2019-10-06
Script for updating our dependencies 


lennet:
It would be nice to have a script to update all our dependencies such as
- bundle update and bundle exec 
- fastlane snapshot update
- license-plist
- swiftformat

mltbnz:
I recently made some good experience with Rake scripts. Could add that

raviTokopedia:
@mltbnz are you working on this? @lennet Are all these actions supposed to be in a single script?

mltbnz:
I didn’t start working on it. Feel free to kick it off. 
If you are starting on this it is probably a good idea to do a draft pull request with the wip so it is visible that somebody is working on this

lennet:
Hi @raviTokopedis, Thank you for contributing!

Yes, one script that triggers all these update commands would be awesome.

Ravi61:
@lennet Added PR #146 for this

PR #146, opened 2019-10-10
Added bash script for Issue #142

Ravi61:
This PR aims to resolve the issue #142 
Name of the script file is `generate.sh`. To use it, just type `./generate.sh`.

### Before merging the PR


generate.sh
```diff
@@ -0,0 +1,34 @@
...
+        bundle update
+    fi
+}
+
+run_swiftformat() {
+    ./bin/swiftformat .
+}
+
+run_license_plist() {
+    ./bin/license-plist
```
Ravi61: @lennet I am unsure as to how to use this license-plist, could you tell me the exact command? Right now, it generates empty plist file.

lennet: Sorry I should have been more precise. We don't wanna run license-plist in swift format with the script. We are already doing that as part of our build phase. We want to update the binary that's stored in the repo. The updates can be find here https://github.com/mono0926/LicensePlist/releases and here https://github.com/nicklockwood/SwiftFormat/releases

Ravi61: Oh ok, in that case, do you think `brew install/upgrade` would be better here? Since, as far as I know, to download from Releases from GitHub, we'd have to do authentication, which becomes tedious if you have two factor auth enabled.

lennet: The latest release for a public repo doesn’t require authentication https://stackoverflow.com/a/54836319

Ravi61: Yup, just found that out! It can be used without auth, but there is a rate limitation, which I don't think should be a problem in this case. 
Have updated the PR to download the releases @lennet, let me know if you have any ideas to make it better!

lennet: Looks awesome! Just one small comment: Lets rename it to something like update_dependencies.sh to make it easier to understand without opening the script itself

Ravi61: Sure, makes sense. Done!

PR #155, opened 2019-10-13
Fix show friends name on map #Cyclehack

huschu:
; disable colliding gesture recognizer in friend view

### DIFF
--- a/CriticalMass/BikeAnnotationView.swift
+++ b/CriticalMass/BikeAnnotationView.swift
@@ -28,8 +28,13 @@ class BikeAnnoationView: MKAnnotationView {
 
 class FriendAnnotationView: MKAnnotationView {
     static let reuseIdentifier = "FriendAnnotationView"
+    var friendView: FriendView?
 
-    var friend: Friend?
+    var friend: Friend? {
+        didSet {
+            friendView?.name = friend?.name ?? ""
+        }
+    }
 
     override init(annotation: MKAnnotation?, reuseIdentifier: String?) {
         super.init(annotation: annotation, reuseIdentifier: reuseIdentifier)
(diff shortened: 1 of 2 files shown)


CriticalMass/ManageFriendsViewController.swift
```diff
@@ -46,9 +46,10 @@ class ManageFriendsViewController: UIViewController, IBConstructable, UITableVie
         tableView.register(cellType: FriendTableViewCell.self)
         tableView.register(cellType: FriendSettingsTableViewCell.self)
         tableView.register(viewType: SettingsTableSectionHeader.self)
-        
-        let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTap))
```
mltbnz: What was this gesture for?

CriticalMass/BikeAnnotationView.swift
```diff
@@ -40,17 +45,18 @@ class FriendAnnotationView: MKAnnotationView {
...
         canShowCallout = false
 
-        let friendView = FriendVie(frame: .init(x: 0, y: 0, width: 109, height: 23))
-        friendView.backgroundColor = .clear
-        friendView.name = friend?.name ?? ""
-        addSubview(friendView)
+        friendView = FriendView(frame: .init(x: 0, y: 0, width: 109, height: 23))
+        friendView!.backgroundColor = .clear
+        friendView!.name = friend?.name ?? ""
+        addSubview(friendView!)
```
mltbnz: Maybe we judt make the FriendView implicitly unwrapped so we can loose a couple of❗️

CriticalMass/BikeAnnotationView.swift
```diff
@@ -40,17 +45,18 @@ class FriendAnnotationView: MKAnnotationView {
...
     }
 }
 
-class FriendVie: UIView {
+class FriendView: UIView {
```
mltbnz: 👍

CriticalMass/ManageFriendsViewController.swift
huschu: It should dismiss the keyboard when tapping anywhere in the table view but it broke the feature where tapping an online friend would focus the friend on the map.

Issue #143, opened 2019-10-06
Dynamic poll rate


lennet:
We are currently updating the Apps content every 12 seconds or after the user send a message as it's using the same API. This may result in an outdated map especially if a request failed because we are currently waiting another 12 seconds for the next try at which point the app might already be in the background. 

It would be nice to be smarter about updating the poll the data every 6 seconds if the app is in foreground and every 12 seconds in background (maybe as well in lower power/ data mode ?). Additionally we should retry sooner, if possible as soon as possible (maybe check for internet connection on iOS 12+ devices https://developer.apple.com/documentation/network/nwpathmonitor)

iglushchuk:
Hi, I would like to make this task, may I?

I have looked at RequestManager, so I think that we can poll the data every 6 seconds and for background update run `application(_:performFetchWithCompletionHandler:)` and BGAppRefreshTask (depends on iOS version)

By the way method `updateData()` might be wrapped to asynchronous operation. It seems that it would be easier to control update, when we have queue.

What do you think, @lennet ?

PR #157, opened 2019-10-14
143 dynamic poll rate

iglushchuk:
#143 I refactored updating data to operation and added operation for waiting for another update request. For update operation I used UIApplication background tasks, to ensure that request will be completed. 

For iOS 12+ I used NWPathMonitor, if we receive path update and It's ok for new network request, we cancel current waiting operation, update operation will fire immediately.
Update interval was changed from 12 to 6 seconds.

I actually did not add tests for cancelling operation with NWPathMonitor, because it's not easy  to directly mock NWPathMonitor, as It is final class, NWPath is struct without initializer. If we really need unit tests for this, I can add protocol and adapter for NWPathMonitor, so I can mock it and call update handlers for test cases by my self.

Looking forward for code review)


CriticalMass/RequestManager.swift
```diff
@@ -120,3 +124,25 @@ public class RequestManager {
...
+            self?.handlePathUpdate(path)
+        }
+        let queue = DispatchQueue(label: "Monitor")
+        pathMonitor.start(queue: queue)
+        
+        pathMonitorRef = pathMonitor
+    }
+
+    func handlePathUpdate(_ path: NWPath) {
+        if path.status == .satisfied {
```
lennet: Isn't this also get called when the network changes from Wifi to cellular?

CriticalMass/RequestManager.swift
```diff
@@ -7,51 +7,78 @@
...
+        queue.maxConcurrentOperationCount = 1
+
+        return queue
+    }()
+
     private var log = OSLog(subsystem: Bundle.main.bundleIdentifier!, category: "RequestManager")
 
-    init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 12.0, idProvider: IDProvider, url: URL) {
+    @available(iOS 12.0, *)
+    convenience init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 6.0, idProvider: IDProvider, url: URL, pathMonitor: NWPathMonitor) {
```
lennet: As you sugested, I would love to see a protocol for NWPathMonitor, this would make testing easier and also remove the second initializer

CriticalMass/RequestManager.swift
```diff
@@ -7,51 +7,78 @@
...
-    }
-
     private struct SendMessagePostBody: Codable {
         var device: String
         var messages: [SendChatMessage]
     }
 
     private let endpoint: URL
 
-    private var hasActiveRequest = false
```
lennet: 👍 Real like that we don't need that anymore :)

CriticalMass/UpdateDataOperation.swift
```diff
@@ -0,0 +1,57 @@
...
+            let request = GetLocationsAndChatMessagesRequest()
+            networkLayer.get(request: request) { [weak self] result in
+                guard let self = self else { return }
+
+                self.result = result
+                self.completeOperation()
+            }
+
+            return
+        }
```
lennet: having a lot of logic in the else case and having two nested guards makes it hard to read for me. Maybe we can find a cleaner solution.

CriticalMass/RequestManager.swift
```diff
@@ -7,51 +7,78 @@
...
     private var log = OSLog(subsystem: Bundle.main.bundleIdentifier!, category: "RequestManager")
 
-    init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 12.0, idProvider: IDProvider, url: URL) {
+    @available(iOS 12.0, *)
+    convenience init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 6.0, idProvider: IDProvider, url: URL, pathMonitor: NWPathMonitor) {
+        self.init(dataStore: dataStore, locationProvider: locationProvider, networkLayer: networkLayer, interval: interval, idProvider: idProvider, url: url)
+        setup(pathMonitor: pathMonitor)
+    }
+
+    init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 6.0, idProvider: IDProvider, url: URL) {
```
lennet: We don't wanna have a constant interval of 6 We should still use 12 if the app is in background

lennet: Additionally we want to use 12 if the app is in low power mode

CriticalMass/RequestManager.swift
iglushchuk: I guess it does get called. When I implement adapter for NWPathMonitor, I will track only status changes

CriticalMass/RequestManager.swift
iglushchuk: In progress

CriticalMass/RequestManager.swift
iglushchuk: I think the same way :)

CriticalMass/RequestManager.swift
iglushchuk: By the way I can refactor LocationManager into operation, not a problem

CriticalMass/AsyncOperation.swift
```diff
@@ -0,0 +1,63 @@
...
+class AsyncOperation: Operation {
+    override var isAsynchronous: Bool {
+        return true
+    }
+    
+    private var _isFinished: Bool = false
+    
+    override var isFinished: Bool {
+        set {
+            willChangeValue(forKey: "isFinished")
```
mltbnz: To track state maybe an enum would be nice:
```swift
enum State: String {
    case ready, executing, finished

    fileprivate var keyPath: String {
      return "is\(rawValue.capitalized)"
    }
  }
```

CriticalMass/AsyncOperation.swift
```diff
@@ -0,0 +1,63 @@
...
+//
+//  Created by Илья Глущук on 13/10/2019.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import Foundation
+
+class AsyncOperation: Operation {
+    override var isAsynchronous: Bool {
+        return true
```
mltbnz: You can omit the return here now 🎉

CriticalMass/AsyncOperation.swift
mltbnz: also then
```swift
var state = State.ready {
  willSet {
    willChangeValue(forKey: newValue.keyPath)
    willChangeValue(forKey: state.keyPath)
  }
  didSet {
    didChangeValue(forKey: oldValue.keyPath)
    didChangeValue(forKey: state.keyPath)
  }
}
```

iglushchuk:
Hi! Sorry, I've been quit busy last days, so I did not have time to continue. Now everything is ok I will start coding this very soon. I mean I did not forget about it :)

CriticalMass/AsyncOperation.swift
iglushchuk: Looks nice, I changed code to your suggestion.

CriticalMass/AsyncOperation.swift
iglushchuk: fixed

CriticalMass/RequestManager.swift
iglushchuk: changed to 12

CriticalMass/UpdateDataOperation.swift
iglushchuk: Agree. Moved code to separate method

iglushchuk:
Hi! I added tests for RequestManager and added protocol for monitoring network changes. Also I refactored LocationManager to work with operation queue the similar way I did with RequestManager. But I did not add test case for it, there is no actually any tests for location manager.

So, I can think about covering LocationManager with tests if it needs. By the way "testIDDoesntChange" for IDStoreTests fails. If you tell me what is the desired logic and behind it I can fix it.

Looking forward for code review :)

lennet:
Thank you! There seem to be some merge conflicts around AppController and RequestManager

iglushchuk:
Branch is up to date with master, no conflicts now.

CriticalMass/AppController.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import UIKit
+import Network
```
lennet: I think this import statement isn't needed anymore

CriticalMass/AppController.swift
```diff
@@ -38,6 +39,14 @@ class AppController {
         return NetworkOperator(networkIndicatorHelper: NetworkActivityIndicatorHelper(), dataProvider: session)
     }()
 
+    private let networkObserver: NetworkObserver? = {
```
lennet: it seems like the networkObserver isn't injected into the RequestManager

CriticalMassTests/RequestManagerTests.swift
```diff
@@ -86,4 +87,32 @@ class RequestManagerTests: XCTestCase {
...
+
+        let exp = expectation(description: "Wait a second")
+        wait(interval: 1) {
+            XCTAssertEqual(setup.networkLayer.numberOfGetCalled, 0)
+            XCTAssertEqual(setup.networkLayer.numberOfPostCalled, 0)
+
+            exp.fulfill()
+        }
+        wait(for: [exp], timeout: 4)
+    }
```
lennet: Nice!

CriticalMass/UpdateDataOperation.swift
```diff
@@ -0,0 +1,62 @@
...
+        
+        let body = SendLocationPostBody(device: idProvider.id, location: currentLocation)
+        guard let bodyData = try? body.encoded() else {
+            state = .finished
+            return
+        }
+
+        let request = PostLocationRequest()
+        networkLayer.post(request: request, bodyData: bodyData) { [weak self] result in
+            guard let self = self else { return }
```
lennet: Shouldn't we set state to finished?

CriticalMass/UpdateDataOperation.swift
```diff
@@ -0,0 +1,62 @@
...
+            self.state = .finished
+        }
+    }
+}
+
+private extension UpdateDataOperation {
+    private func getLocationsAndChatMessages() {
+        let request = GetLocationsAndChatMessagesRequest()
+        networkLayer.get(request: request) { [weak self] result in
+            guard let self = self else { return }
```
lennet: same here

CriticalMass/LocationManager.swift
```diff
@@ -24,6 +24,13 @@ class LocationManager: NSObject, CLLocationManagerDelegate, LocationProvider {
         }
     }
 
+    private let operationQueue: OperationQueue = {
```
lennet: Having a separate OperationQueue for the LocationManager is problematic because the Location might be outdated when the request is happening as both operation queues might be out of sync. We should refactor the requesting and getting a location in an Async Operation and integrate it into the RequestManagers Operation queue

CriticalMassTests/RequestManagerTests.swift
```diff
@@ -9,13 +9,14 @@
 import XCTest
 
 class RequestManagerTests: XCTestCase {
```
lennet: Can you add one more test case for recovering after regaining network access?

CriticalMass/UpdateDataOperation.swift
iglushchuk: I set state to finished in completion block after result was assigned. Do you think we should set finished before assignment?

CriticalMass/AppController.swift
iglushchuk: My fault, fixed

CriticalMass/AppController.swift
iglushchuk: right, fixed

CriticalMass/LocationManager.swift
iglushchuk: I think I understand the problem, I need a little time to figure out how to refactor LocationManager code.

CriticalMass/UpdateDataOperation.swift
iglushchuk: If we are talking about "return" in guard, then self doesn't exists there. It means that operation queue was deallocated (with request manager) and I think we don't need to set state to finished there. By the way there is compilation error: "Variable declared in 'guard' condition is not usable in its body".

iglushchuk:
Hi! I added test for regaining network access and refactored updating location to async operation. So, looking forward for review, I'm ready for discussion)

CriticalMassTests/RequestManagerTests.swift
```diff
@@ -86,4 +87,54 @@ class RequestManagerTests: XCTestCase {
...
+        setup.networkObserver.update(with: .none)
+
+        let exp = expectation(description: "Wait a second")
+        wait(interval: 1) {
+            XCTAssertEqual(setup.networkLayer.numberOfGetCalled, 0)
+            XCTAssertEqual(setup.networkLayer.numberOfPostCalled, 0)
+
+            setup.networkObserver.update(with: .satisfied)
+            self.wait(interval: 1) {
+                setup.networkObserver.update(with: .satisfied)
```
lennet: shouldn't be one update call be enough?

CriticalMass/UpdateDataOperation.swift
lennet: oh makes sense 👍

CriticalMass/LocationProvider.swift
```diff
@@ -17,4 +17,8 @@ public enum LocationProviderPermission: String {
 public protocol LocationProvider {
     var currentLocation: Location? { get }
     static var accessPermission: LocationProviderPermission { get }
+
+    func updateLocation()
+    var locationUpdateHandler: (() -> Void)? { get set }
+    var locationErrorHandler: (() -> Void)? { get set }
```
lennet: Minor: maybe we can pass a completion block into updateLocation instead of exposing the blocks publicly. We also could use Result here similar to ResultCallback

CriticalMass/UpdateDataOperation.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    override func main() {
+        guard let currentLocation = locationProvider.currentLocation else {
+            getLocationsAndChatMessages()
+            return
+        }
+
+        let body = SendLocationPostBody(device: idProvider.id, location: currentLocation)
+        guard let bodyData = try? body.encoded() else {
+            state = .finished
```
lennet: We also need a result here

CriticalMass/RequestManager.swift
```diff
@@ -15,46 +15,69 @@ import os.log
...
+        let queue = OperationQueue()
+        queue.maxConcurrentOperationCount = 1
+
+        return queue
+    }()
 
     private var log = OSLog(subsystem: Bundle.main.bundleIdentifier ?? "", category: "RequestManager")
 
-    public init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 12.0, idProvider: IDProvider) {
+    public init(dataStore: DataStore, locationProvider: LocationProvider, networkLayer: NetworkLayer, interval: TimeInterval = 12.0, idProvider: IDProvider, networkObserver: NetworkObserver?) {
```
stephanlindauer: so 12 remains the default?

CriticalMass/LocationProvider.swift
iglushchuk: Nice idea, I changed updateLocation() to method with completion block

CriticalMassTests/RequestManagerTests.swift
iglushchuk: yeah, removed unnecessary update

CriticalMass/RequestManager.swift
iglushchuk: I left 12 seconds as it was before refactoring, can you please explain more about 1 second interval? I mean I understand that frequent requests might be a problem for server, but how user can schedule requests every second, if we set it in code for 12 seconds? Do you thing we should increase intervals? Or there should be something like minimum waiting interval?

CriticalMass/UpdateDataOperation.swift
iglushchuk: ok, added

CriticalMass/RequestManager.swift
stephanlindauer: sorry maybe i misunderstood. from the title of the pr i thought this should enable the user to dynamically set the poll interval.

iglushchuk: Actually no, we don't allow users to set poll interval. The main goal sounds like: if user lost internet connection and regain it later, we retrieve location and send update request as soon as possible. But next request will be send with poll interval of 12 seconds.

stephanlindauer: :+1:

CriticalMass/WaitOperation.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  WaitOperation.swift
+//  CriticalMaps
+//
+//  Created by Илья Глущук on 13/10/2019.
+//  Copyright © 2019 Pokus Labs. All rights reserved.
+//
+
+import UIKit
```
lennet: Replacing Import UIKit with Import foundation should fix some of the SPM errors

CriticalMass/RequestManager.swift
```diff
@@ -15,46 +15,69 @@ import os.log
...
     }
 
-    @objc private func timerDidUpdate(timer _: Timer) {
-        Logger.log(.info, log: log, "Timer did update")
-        updateData()
+    private func addUpdateOperation(with interval: TimeInterval) {
+        let updateDataOperation = UpdateDataOperation(locationProvider: locationProvider,
+                                                      idProvider: idProvider,
+                                                      networkLayer: networkLayer)
+        let taskIdentifier = UIApplication.shared.beginBackgroundTask {
```
lennet: We might need to. add #if canImport(UIKit) for parts that uses UIKit Apis for  SPM compatibility

CriticalMass/WaitOperation.swift
iglushchuk: fixed, check it please

CriticalMass/RequestManager.swift
iglushchuk: fixed

CriticalMass/RequestManager.swift
```diff
@@ -128,3 +132,18 @@ public class RequestManager {
         }
     }
 }
+
```
lennet: The extensions also needs to be marked as @available(macOS 10.12, *) for spm compatibility

iglushchuk: fixed

lennet:
Yeah 🙌! Thanks for your patience with the change requests and finals spm support stuff

iglushchuk:
Thanks for interesting issue :)

PR #172, opened 2019-10-22
Move Chat Presentation to viewDidLoad

### LINKED ISSUE #171: Bug: Swipe Down Gesture in Modal changes Content
**Current Behaviour**
When you swipe a bit down in the `SocialViewController`, but don't dismiss the modal with the swipe down gesture, the content of the view changes but the selected segmented control index stays the same. 

dismiss

**Expected Behaviour**
When you don't dismiss the modal, the content should stay the same.

fbernutz:
Fixes #171 

This moves the call from `viewWillAppear` to `viewDidLoad`, because when you slightly pull down the modal with the swipe gesture `viewWillAppear` gets called and changes the content. Everytime the modal is presented `viewDidLoad` is called, so I think this is the better place to setup the content. 

dismiss-correct

### Before merging the PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Changelog for Critical Maps iOS
 
+## [Unreleased]
+
+### Fixed
+
+- Fix: Don't update content when slightly swipe down in Social Modal
+
 ## [3.4.0] - 2019-10-08
 
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ Changelog for Critical Maps iOS
 - Fix: NavigationBar Colors under iOS 13 
 - Fix: UITableViewHeaderFooterView backgroundColor deprecation warning
 - Fix: Ambiguous auto layout constraints for Settings screen
+- Fix: Don't update content when slightly swipe down in Social Modal
```
fbernutz: I don't know if this is the correct line for this or if I should put it under a new version on top?

lennet: cc @headione

mltbnz: I would say you can open an Unreleased section since a new version will be > 3.5, right?!
https://keepachangelog.com/en/1.0.0/

fbernutz: Okay, done ✅

PR #177, opened 2019-10-22
Fixed UITests on Travis

lennet:
Travis Xcode version doesn't has the iPhone SE installed, which breaks our UITests. Removing SE from the snap file in case it's not installed fixes the issue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,5 @@ jobs:
   include:
     - script: bundle exec fastlane test
       name: UnitTests
-    - script: bundle exec fastlane screenshots_limited
+    - script: bundle exec fastlane ui_test
       name: UITests
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -41,6 +41,18 @@ platform :ios do
       scheme: Scheme_UnitTests
     )
   end
+
+
+  desc 'Run UITests'
+  lane :ui_test do
+    sh "cd .. && xcodebuild -resolvePackageDependencies"
+
+    run_tests(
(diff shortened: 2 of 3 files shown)


fbernutz:
Maybe it helps... I ran into the same issue in one of my projects.
It is weird that you can't configure `capture_ios_screenshots` to not use the `Snapfile` configuration. Because we don't want to use the "iPhone SE" in the CI at all, but I think fastlane just checks if there is a `Snapfile` and want to use these devices and after that check it uses the configuration (see `screenshots_limited`) from the `Fastfile`. But this doesn't happen because it fails when searching the "iPhone SE". 
So what I did... I deleted the `Snapfile` completely and moved the device and language configuration into the actions `screenshots_complete` and `screenshots_limited` in the Fastfile. An advantage is that the pipeline is green again, because we don't use the "iPhone SE" in the CI and it can't fail when it's not installed, but a disadvantage is that the `Fastfile` is getting kind of big with configurations you normally would expect in a `Snapfile`.

lennet:
Good point. I added a new lane that just runs the UITest without screenshot capturing to unblock us for now. Especially since we currently don't have a way to extract the screenshot from the CI, so we don't need to capture them anyway. 

Would be cool to reenable that in the future though, especially if we can somehow automatically attach the generated Screenshots to a PR.

lennet:
Looks like our CI isn't currently running any tests 👀

+--------------------+---+
|      Test Results      |
+--------------------+---+
| Number of tests    | 0 |
| Number of failures | 0 |
+--------------------+---+

PR #178, opened 2019-10-22
Fix whitespace only chat message and rework chatInputView a little

### LINKED ISSUE #173: Chat should not be able to send whitespace
My idea for the fix: Adding whitespace in the chat should enable the send button but if only whitespace has been added the send event should not trigger the post request.

mltbnz:
* renaming for clarity
* using layout anchors
* gate send event with check if text isEmpty and does not only contain of whitespace

### Before merging the PR

closes #173

### DIFF
--- a/CriticalMass/ChatInputView.swift
+++ b/CriticalMass/ChatInputView.swift
@@ -14,6 +14,8 @@ protocol ChatInputDelegate: AnyObject {
 class ChatInputView: UIView, UITextFieldDelegate {
     struct Constants {
         static let textFieldHeight = CGFloat(46.0)
+        static let sendButtonWidth = CGFloat(65.0)
+        static let messageTextFieldInsets = UIEdgeInsets(top: 0, left: 16, bottom: 0, right: 16)
     }
 
     @objc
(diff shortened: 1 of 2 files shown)


CriticalMass/ChatInputView.swift
```diff
@@ -12,25 +12,29 @@ protocol ChatInputDelegate: AnyObject {
 }
 
 class ChatInputView: UIView, UITextFieldDelegate {
+    struct Constants {
+        static let textFieldHeight = CGFloat(46.0)
```
lennet: Does that work with dynamic type?

CriticalMass/ChatInputView.swift
```diff
@@ -70,11 +83,10 @@ class ChatInputView: UIView, UITextFieldDelegate {
     }
 
     private func commonInit() {
-        addSubview(textField)
-        addSubview(button)
+        addSubview(chatStack)
         addSubview(separator)
 
-        textField.delegate = self
+        messageTextField.addTarget(self, action: #selector(didTapSendButton), for: .editingDidEndOnExit)
```
lennet: Cool! Never seen that API before

CriticalMass/ChatInputView.swift
mltbnz: That is a good question. I have not tried tbh. Never saw out in the os but I always thought why is no search bar or whatsoever updated with a bigger font. I will take a look in what is possible there

lennet: PNG-Bild
Just checked iMessage and the text input gets updated. But if just seen that we are not supporting it in the current version of the app so its finde that this PR isn‘t changing that. We just should make sure that nothing is broken but it doesn’t need to scale for now

mltbnz: Also safari supports it in the url textfield. Maybe we can create an issue for it. Would be cool if we support it

mltbnz:
I figured that the PR has an issue when entering a lot of text. I am not setting a sendButton width constraint and that makes the send Button disappear. When I add a constraint works nice but I figured that longer strings like in french make textField very small especially on smaller devices. So maybe we could switch from a string button to an icon button.
What do you think @zutrinken ?

lennet:
Maybe setting a maxwidth and truncate the text if needed?

mltbnz:
I have set a fixed width that works with our localisations. The finnish translation `lähettää` for example would be truncated

zutrinken:
@mltbnz ja, let's do button like iMessage. I'll update the design.

zutrinken:
@mltbnz does this work for you?

<table><tr><td></td><td></td></tr></table>

mltbnz:
I like it

PR #184, opened 2019-10-29
Merge some of the Friends feature Branch into Master

lennet:
This PR contains some of the Changes from the Friends Feature Branch without changing any functionality. Splitting the PR into multiple PRs makes it easier to review. 

Most of the code introduced with this PR doesn't get called without the rest of the Friends feature branch. The biggest change is that the SwiftHash framework has been replaced with Crypto


CriticalMass/IBConstructable.swift
```diff
@@ -32,6 +32,18 @@ extension IBConstructable where Self: UIViewController {
...
+        return typeName
+    }
+
+    static var bundle: Bundle {
+        return Bundle(for: Self.self)
+    }
+
+    static var nib: UINib {
+        return UINib(nibName: nibName, bundle: bundle)
+    }
```
mltbnz: yeah. wanted that on viewcontrollers a couple of times already

CriticalMass/UIView+Autolayout.swift
```diff
@@ -0,0 +1,22 @@
...
+
+extension UIView {
+    func addLayoutsCenter(in view: UIView, size: CGSize) {
+        translatesAutoresizingMaskIntoConstraints = false
+
+        view.addConstraints([.init(item: self, attribute: .centerX, relatedBy: .equal, toItem: view, attribute: .centerX, multiplier: 1, constant: 1),
+                             .init(item: self, attribute: .centerY, relatedBy: .equal, toItem: view, attribute: .centerY, multiplier: 1, constant: 0)])
+        addConstraints([
+            .init(item: self, attribute: .height, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: size.height),
+            .init(item: self, attribute: .width, relatedBy: .equal, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: size.width),
```
mltbnz: I would use the anchor properties of the views to build the comstraints because to me they remove noise like the multiplier which is not used here

CriticalMass/IBConstructable.swift
mltbnz: Do we want to go Swift 5.1 and omit the return statements? I also had to confirm UIviewController to the TypeNameProtocol

CriticalMass/Location.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-struct Location: Equatable {
+public struct Location: Equatable, Hashable {
```
mltbnz: Isn’t just Hashable enough since it implements Equatable I think

CriticalMass/TypeNameProtocol.swift
```diff
@@ -24,3 +24,4 @@ public extension TypeNameProtocol {
 }
 
 extension UIView: TypeNameProtocol {}
+extension UIViewController: TypeNameProtocol {}
```
mltbnz: Ah. Hey it is

CriticalMass/Location.swift
lennet: Good point!

PR #185, opened 2019-10-29
Fix action indicator visibility in Dark Mode

### LINKED ISSUE #175: The action indicator is hardly visible in Dark Mode
Observed:
image

This is how it should look like

Bildschirmfoto 2019-10-22 um 19 11 35

More infos in the design repo  http://github.com/criticalmaps/criticalmaps-design

dvrzan:

Fixes #175 
Added a small workaround for activity indicator visibility in Dark Mode. Fixed by adding a button and setting the color, instead of system Disclosure Indicator.

### Before merging the PR

Activity-indicator-in-dark-mode-screenshot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for Critical Maps iOS
 
 - Fix: Don't update content when slightly swipe down in Social Modal
 - Fix whitespace only chat message can not be send anymore.
+- Fix: Action indicator visibility in Dark Mode
 
 ## [3.4.0] - 2019-10-08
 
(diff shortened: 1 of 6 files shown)


lennet:
Thanks for the PR! Thats indeed fixing the bug but I would prefer a solution that doesn't require a custom button if possible.

Have you tried solutions that doesn't require adding a custom button? Overriding the UserInterfaceStyle in the SettingsViewController (similar to what we do in MapsViewController.configureTileRenderer) might work

dvrzan:
No problem. I will try to make it work by overriding the UserInterfaceStyle. Thank you for the feedback!

mltbnz:
Is this one still in work?

PR #187, opened 2019-11-01
Fix Dark Mode for Rules Detail

### LINKED ISSUE #174: Rules Detail view isn't readable in Dark mode
image_from_ios

This is a regression probably caused by #132 

This is how the text color should look like 
Bildschirmfoto 2019-10-22 um 19 08 01

more details in the design repo http://github.com/criticalmaps/criticalmaps-design

We are currently using UITextView with attributed attachments which makes the configuration overly complicated. Maybe worth migrating to a UIStackView and a plain UIImageView + UILabel while fixing the bug

fbernutz:
Fixes #174 

The `RulesDetailViewController` now uses a `UIStackView` with a `UIImageView` and `UILabel` instead of an `attributedString`. I added a scroll view, so it also supports dynamic type now. 

### Before 

```
RulesDetailViewController
└── UITextView
```

A Screenshot is already in #174.

### After

```
RulesDetailViewController
└── UIScrollView
    └── UIStackView
        └── UIImageView
        └── UILabel
```

| Dark | Light |
|---|---|
| Simulator Screen Shot - iPhone 11 Pro - 2019-11-01 at 13 50 33 | Simulator Screen Shot - iPhone 11 Pro - 2019-11-01 at 13 50 42 |

| Landscape Dark | Landscape Light |
|---|---|
| Simulator Screen Shot - iPhone 11 Pro - 2019-11-02 at 15 04 41 | Simulator Screen Shot - iPhone 11 Pro - 2019-11-02 at 15 04 13 |

| Landscape Dark Dynamic Type | 
|---|
| Simulator Screen Shot - iPhone 11 Pro - 2019-11-02 at 15 06 49 | 

### Before merging the PR


fbernutz:
I'm not sure why, but the textColor only changes to light (in dark mode) in iOS 13. All other iOS versions (iOS 10, 11, 12), the text is dark... Do you have an idea how to fix this? I use the following to set the text color: 
```swift
UILabel.appearance(whenContainedInInstancesOf: [RulesDetailViewController.self]).textColor = theme.titleTextColor
```

fbernutz:
> I'm not sure why, but the textColor only changes to light (in dark mode) in iOS 13. All other iOS versions (iOS 10, 11, 12), the text is dark... Do you have an idea how to fix this? I use the following to set the text color:
> 
> ```swift
> UILabel.appearance(whenContainedInInstancesOf: [RulesDetailViewController.self]).textColor = theme.titleTextColor
> ```

I fixed it with adding a new `RuleDescriptionLabel` subclass... now it works for all iOS versions.

CriticalMass/RulesDetailViewController.swift
```diff
@@ -7,19 +7,52 @@
...
+        return imageView
+    }()
+
+    private lazy var descriptionLabel: RuleDescriptionLabel = {
+        let label = RuleDescriptionLabel()
+        label.text = rule.text
+        return label
+    }()
+
+    private lazy var imageViewHeightConstraint = { () -> NSLayoutConstraint in
```
lennet: Minor: how about `private lazy var imageViewHeightConstraint: NSLayoutConstraint = { ` to be consistent with the other lazy properties

CriticalMass/RulesDetailViewController.swift
```diff
@@ -7,19 +7,52 @@
...
+        let aspectRatio: CGFloat
+        if let artwork = rule.artwork {
+            aspectRatio = artwork.size.height / artwork.size.width
+        } else {
+            aspectRatio = 0.6
```
lennet: shouldn't it be 0 if we don't have an artwork?

zutrinken:
> Maybe we can change the orientation of the stack view in landscape mode and show the artwork + text side by side.

Good idea!

@fbernutz You need a design or just try it?

fbernutz:
> > Maybe we can change the orientation of the stack view in landscape mode and show the artwork + text side by side.
> 
> Good idea!
> 
> @fbernutz You need a design or just try it?

I think I'm fine without a design :) Thanks for reviewing, I totally forgot about landscape mode 🙈

CriticalMass/RulesDetailViewController.swift
fbernutz: of course, fixed it

CriticalMass/RulesDetailViewController.swift
fbernutz: 👍

fbernutz:
I added some screenshots in the PR description. Please have a look if this is what you've imagined :)

fbernutz:
By the way, have you already thought about moving the rules into a [UIPageViewController](https://developer.apple.com/documentation/uikit/uipageviewcontroller)? I think about that the use case is, that a user wants to read all the rules and not a specific rule. Right now, the user has to tap on a rule to read about it, then swipe to go back and tap on the next rule, and so on. If you are interested, I could open an issue for this. I'm not sure how it's handled in the android app and if you want to stay consistent.

zutrinken:
@fbernutz the screenshots are looking good. I could imagine a whole new concept where you swipe through the rules for example. But that needs some more concept work and I personally would tackle this after we got better illustrations and icons for all rules.

Issue #141, opened 2019-10-06
Critical Mass simulation for debugging 


lennet:
We are currently don't have a dedicated environment for testing are directly hitting the production backend for debugging and testing. This works for most use cases.

However, some scenarios require a lot of activity on the map. If would be nice if we could mock the data that the App is using locally for test purposes or even replay recorded backend responses. 

Testdata can be found here https://github.com/criticalmaps/criticalmaps-testdata

PR #188, opened 2019-11-02
Introduced NetworkDataProvider and added tests for NetworkOperator

lennet:
Initially the NetworkOperator was just an abstraction for URLSessions but it gained some logic over the time so I introduced another abstraction layer to be able to mock network request without replacing a lot of core logic.

This allows us to write better test and read data from files for debugging purposes instead from a network connection (related to #141 ).

Test coverage for UnitTests increased from 26% to 30%

We could think about removing the NetworkLayer protocol as it isn't needed anymore because  test mocking is also possible with injecting a custom NetworkDataProvider.


CriticalMass/Constants.swift
```diff
@@ -8,7 +8,6 @@
 import Foundation
 
 enum Constants {
-    static let twitterEndpoint = URL(string: "https://api.criticalmaps.net/twitter")!
```
lennet: This is not really related to the PR. Just some small cleanup of clean code

CriticalMass/NetworkOperator.swift
```diff
@@ -57,7 +46,7 @@ struct NetworkOperator: NetworkLayer {
     private func dataTask(with request: URLRequest,
                           completion: @escaping ResultCallback<Data>) {
         networkIndicatorHelper.didStartRequest()
```
lennet: The new tests aren't covering networkIndicator activation. Not sure if tests for that are needed though

CriticalMass/AppController.swift
```diff
@@ -26,11 +26,15 @@ class AppController {
...
 
     private lazy var requestManager: RequestManager = {
-        RequestManager(dataStore: dataStore, locationProvider: LocationManager(), networkLayer: networkOperator, idProvider: idProvider, url: Constants.apiEndpoint)
+        RequestManager(dataStore: dataStore, locationProvider: LocationManager(), networkLayer: networkOperator, idProvider: idProvider)
     }()
 
     private let networkOperator: NetworkOperator = {
-        NetworkOperator(networkIndicatorHelper: NetworkActivityIndicatorHelper())
+        let configuration = URLSessionConfiguration.default
+        configuration.requestCachePolicy = .reloadIgnoringLocalCacheData
```
mltbnz: While adding an error view I thought we might need to set a timeout timeinterval as well. But since it is not related to this PR not requested here. Just thought we could discuss here

CriticalMass/AppController.swift
```diff
@@ -44,7 +48,7 @@ class AppController {
     }()
 
     private lazy var twitterManager: TwitterManager = {
-        TwitterManager(networkLayer: networkOperator, url: Constants.twitterEndpoint)
+        TwitterManager(networkLayer: networkOperator, request: TwitterRequest())
```
mltbnz: good idea

CriticalMass/AppController.swift
lennet: Doesn’t the default configuration come with a timeout?

lennet:
> Looks good. Might cause some trouble in my ErrorState PR but selber schuld :D

I’m okay with waiting with this PR until the ErrorState one got merged

CriticalMass/AppController.swift
mltbnz: No. E.g the Twitter Feed was loading all the time when the BE was not responding

PR #192, opened 2019-11-06
add send icon for chat input view

### LINKED ISSUE #183: Make chat input respond to dynamic type
To make the App more accessible the chat input view font size and the size of the textfield should react to system font size changes. For example the messages app on iOS 

146A2905-8FF9-4C79-81C7-D0BB5BC61075

### LINKED ISSUE #186: Use Icon instead of text for send button in Chat
<table><tr><td></td><td></td></tr></table>

_Originally posted by @zutrinken in https://github.com/criticalmaps/criticalmaps-ios/pull/178#issuecomment-547627974_

mltbnz:
Changes the send action button in the ChatInputView

closes #186
closes #183  

Screenshot 2019-11-06 at 21 49 26
Screenshot 2019-11-06 at 21 49 40
Screenshot 2019-11-06 at 21 49 52
Screenshot 2019-11-06 at 21 49 58

Also will add GrowingTextView to make the input more accessable
Screenshot 2019-11-17 at 14 16 17

And replaces the loading indicator while sending a chat message with the CM LoadingIndocator 
Screenshot 2019-11-17 at 14 21 24

### Before merging the PR


CriticalMass/Images.xcassets/icon-send.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
...
+      "filename" : "icon-send.pdf"
+    }
+  ],
+  "info" : {
+    "version" : 1,
+    "author" : "xcode"
+  },
+  "properties" : {
+    "template-rendering-intent" : "template",
+    "preserves-vector-representation" : true
```
lennet: Cool! Though this is only supported with iOS 11 +

CriticalMass/ChatInputView.swift
```diff
@@ -11,109 +11,54 @@ protocol ChatInputDelegate: AnyObject {
     func didTapSendButton(text: String)
 }
 
-class ChatInputView: UIView, UITextFieldDelegate {
-    struct Constants {
-        static let textFieldHeight = CGFloat(46.0)
-        static let sendButtonWidth = CGFloat(65.0)
+class ChatInputView: UIView, IBConstructable {
```
lennet: The sendButton needs an accessibilityLabel

CriticalMass/ChatInputView.swift
```diff
@@ -11,109 +11,54 @@ protocol ChatInputDelegate: AnyObject {
...
-            sendButton.setTitleColor(newValue?.withAlphaComponent(0.4), for: .highlighted)
+            sendButton.tintColor = newValue
+        }
+    }
+    @objc
+    dynamic var sendMessageButtonBGColor: UIColor? {
+        willSet {
+            sendButton.backgroundColor = newValue
+            var highLightBackgroundColor: UIColor {
+                return newValue == .white ? .lightGray : .darkGray
```
lennet: Comparing UIColor objects might break two UIColors might result in the same color but not be equal UIColor objects. Adding a specific color for the state might be safer and easier to maintain

CriticalMass/ChatInputView.xib
```diff
@@ -0,0 +1,85 @@
...
+                            <rect key="frame" x="342" y="3" width="40" height="40"/>
+                            <constraints>
+                                <constraint firstAttribute="height" constant="40" id="Lt4-S8-mss"/>
+                                <constraint firstAttribute="width" secondItem="ok1-Cv-1uN" secondAttribute="height" multiplier="1:1" id="Obz-Yv-vIp"/>
+                            </constraints>
+                            <color key="tintColor" systemColor="labelColor" cocoaTouchSystemColor="darkTextColor"/>
+                            <state key="normal" image="icon-send"/>
+                            <userDefinedRuntimeAttributes>
+                                <userDefinedRuntimeAttribute type="number" keyPath="cornerRadius">
+                                    <real key="value" value="20"/>
```
lennet: This works for now but will break with #183

CriticalMass/ChatInputView.swift
```diff
@@ -139,3 +91,17 @@ private extension String {
         !self.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
     }
 }
+
+private extension UIButton {
+    func setBackgroundColor(color: UIColor, forState: UIControl.State) {
```
lennet: I don’t really like that solution. If there is no other way we should use the UIGraphicsImageRenderer api. 
Maybe a similar to approach to CustomButton.swift might be an optio.

CriticalMass/ChatInputView.swift
mltbnz: I knew this would come back at me 😀 I would like an opinion from @zutrinken which grey would fit for both themes

CriticalMass/ChatInputView.swift
mltbnz: Was the simplest one without subclassing UIButton. Would you draw a shape?

CriticalMass/Images.xcassets/icon-send.imageset/Contents.json
mltbnz: Another case for bumping the target up maybe

lennet: I tested it with the iOS 10 simulator and it seemed to work

mltbnz: Thanks for testing

CriticalMass/ChatInputView.swift
lennet: I would invert the shape of the pdf and just use the default behavior that a UIButton offers when it’s marked as disabled

CriticalMass/ChatInputView.xib
mltbnz: Do you mean the IB solution?

lennet: using a hardcoded corner radius for a circular shape

CriticalMass/ChatInputView.swift
mltbnz: maybe you can post some example code here

lennet: Group
 basically I would suggest using that shape as the button image (the arrow should be transparent). That would solve the issue around corner radius for changing frames + give us the default disabled behaviour without any custom logic

mltbnz: Ok. Then I think we need an asset update. I think the current approach fits in the theming control. Now I moved the cornerRadius in the layoutSubviews

CriticalMass/ChatInputView.xib
mltbnz: right. moved that into the view

CriticalMass/ChatInputView.swift
mltbnz: Giving it a second thought I don’t think we should base the button implementation on the asset. I’m in favor of having a iconbutton class maybe. The button will not grow in size increasing dynamic type anyways

mltbnz:
Sure. I will then also rework the logic into a controller because there is just too much going on for a dumb view imho. Will also try my luck with replacing the textField with a textView because of #183

CriticalMass/ChatInputViewController.xib
```diff
@@ -0,0 +1,90 @@
...
+                                    <color key="value" white="0.66666666666666663" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                                </userDefinedRuntimeAttribute>
+                            </userDefinedRuntimeAttributes>
+                            <connections>
+                                <outlet property="delegate" destination="-1" id="lNh-fE-Jtu"/>
+                            </connections>
+                        </textView>
+                        <button opaque="NO" clipsSubviews="YES" contentMode="scaleToFill" enabled="NO" contentHorizontalAlignment="center" contentVerticalAlignment="center" adjustsImageSizeForAccessibilityContentSizeCategory="YES" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="ok1-Cv-1uN" customClass="SendButton" customModule="CriticalMaps" customModuleProvider="target">
+                            <rect key="frame" x="352" y="4" width="38" height="38"/>
+                            <accessibility key="accessibilityConfiguration" hint="Sends the chat message" label="send message button"/>
```
lennet: the accessibilityLabel shouldn't contain the elements trait "button" as this gets already announced by VoiceOver and it needs to be translated

lennet: just "send" should be enough as a Label

CriticalMass/GrowingTextView.swift
```diff
@@ -0,0 +1,190 @@
+//
+//  GrowingTextView.swift
+//
+//  Created by Kenneth Tsang on 17/2/2016.
+//  Copyright (c) 2016 Kenneth Tsang. All rights reserved.
+//
+// https://github.com/KennethTsang/GrowingTextView
```
lennet: GrowingTextView is licensed under MIT which requires attribution in the App

CriticalMass/MessagesTableViewController.swift
```diff
@@ -36,12 +36,17 @@ class MessagesTableViewController<T: IBConstructableMessageTableViewCell>: UITab
...
+    }
+
+    private func setupTableView() {
         // Setting the footerView hides seperators for empty cellls
         tableView.tableFooterView = UIView()
         tableView.register(cellType: T.self)
         // To use UITableViews dynamicHeight
         tableView.rowHeight = UITableView.automaticDimension
         tableView.estimatedRowHeight = 110.0
+        tableView.separatorColor = .gray300
```
lennet: Why do we need a custom color here. Does that work with light/ dark mode?

CriticalMass/ChatViewController.swift
```diff
@@ -7,24 +7,32 @@
...
+        NSLayoutConstraint(item: chatInput.view!, attribute: .bottom, relatedBy: .equal, toItem: view, attribute: .bottom, multiplier: 1, constant: 0)
     }()
 
     private lazy var chatInputHeightConstraint = {
-        chatInput.heightAnchor.constraint(equalToConstant: Constants.chatInputHeight)
+        NSLayoutConstraint(item: chatInput.view!, attribute: .height, relatedBy: .lessThanOrEqual, toItem: nil, attribute: .notAnAttribute, multiplier: 1, constant: Constants.chatInputHeight)
     }()
 
-    init(chatManager: ChatManager) {
+    init(chatManager: ChatManager, chatInputViewController: ChatInputViewController) {
```
lennet: What are the benefits here passing in the ChatInputViewController with the initialiser? Maybe create it within the ChatViewController?

CriticalMass/ChatInputViewController.xib
mltbnz: Thanks. I am a little unsure what is the best practice here

CriticalMass/MessagesTableViewController.swift
mltbnz: Works for both. 
I took a look the design and updated it

CriticalMass/ChatViewController.swift
mltbnz: I did before but the input requires the themeController now since I can not set the keyboard appearance via uiappearance for the textView and since we injected that before I tought I should do the same here for consitancy. TBH the themeController approach seems not very practical anymore but maybe I can do a refactor PR next

CriticalMass/GrowingTextView.swift
mltbnz: Should I add the license? Does not SPM and I just dragged it in

mltbnz:
@zutrinken Was not really practical for now to put the sendButton inside the textView so the current design looks a little different than in Zeplin

CriticalMass/ChatViewController.swift
lennet: We could inject the ThemeController for now into the ChatViewController to create the ChatInputViewController inside the ChatViewController to keep it out of the AppController

CriticalMass/GrowingTextView.swift
lennet: Yes, it requires a license in the App. Not sure if license-plist supports multiple license sources though

CriticalMass/ChatViewController.swift
mltbnz: Fine with me. What if I say I am thinking  a themController Singleton? Think for testing it should still do the job

CriticalMass/GrowingTextView.swift
mltbnz: Just adding the license file could work since it is looking for those I think

CriticalMass/ChatViewController.swift
lennet: It might to the job right now but it might become painful in the future. For now I would prefer injecting the theme controller into the view controller that need it

lennet:
Is the accessibility label for the send button localised?

CriticalMass/MessagesTableViewController.swift
lennet: have you tried overriding the userinterfacestyle depending on the current theme? That should fix the separator color without defining a custom one

mltbnz: haven't tried that since it is only available in iOS 12 ⬆️ I think

mltbnz:
> Is the accessibility label for the send button localised?

Good point. We actually have that one localised. Will add it

mltbnz:
Adding it via yaml didn’t work and the PR for SPM support on the repo does not move. I actually added two lines to the code so I would assume it is like in art where an altered picture for example becomes a new peace of art 😄 Not sure what to do with it otherwise

lennet:
😄 We still need to attribute the copyright holder. There are some possible solutions to fix issue
1. Fork GrowingTextView and add SPM support and use that fork until the main package supports it
2. Remove license-plist from build phase and manually update the bundle
3. Contribute to license-plist to add support licenses from yaml additionally to SPM

mltbnz:
Sure there are options. I go with the fork then I guess since it seems the most reasonable. Could also point the package to the commit hash I guess. Will find a way 🗺

mltbnz:
> Sure there are options. I go with the fork then I guess since it seems the most reasonable. Could also point the package to the commit hash I guess. Will find a way 🗺

Forked it and added it to the project

PR #193, opened 2019-11-10
Mass simulation

lennet:
This PR implements #141 and consists of 3 parts

1. Making the core logic of the app reusable as an Swift Package
2. Use 1. to create a script that collects snapshots
3. Add option to replace the Default NetworkDataProvider with local snapshots

## Todos:
- [ ] More documentation
- [x] Add default snapshots
- [x] Add build scheme with simulation enabled
- [x] Add `swift build` to travis 

Happy for Feedback! Especially about exposing the logic as an Swift Package as it adds some noise to our codebase but on the other enables us to collect snapshots under realistic conditions


CriticalMass/AppController.swift
```diff
@@ -8,34 +8,27 @@
...
-
-    public func onWillEnterForeground() {
-        if #available(iOS 10.3, *) {
-            RatingHelper().onEnterForeground()
-        }
-    }
-
     private var idProvider: IDProvider = IDStore()
     private var dataStore = MemoryDataStore()
+    private var simulationModeEnabled =   false
```
mltbnz: evil whitespace

PR #195, opened 2019-11-16
MapViewController Refactoring

lennet:
I moved Annotation specific logic out of MapViewController into its own controller as the current MapView infrastructure becomes hard to maintain if we add more logic to it in the future. 

If we wanna add more annotationTypes (Friends, events etc), we'll just need to create a new `AnnotationController` subclass  and add that to `private lazy var annotationController: [AnnotationController]` and the rest should work out of the box

### DIFF
--- a/CriticalMass/BikeAnnotationController.swift
+++ b/CriticalMass/BikeAnnotationController.swift
@@ -12,8 +12,6 @@ class BikeAnnotation: IdentifiableAnnnotation {}
 
 class BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoationView> {
     public override func setup() {
-        mapView.register(annotationType: BikeAnnoationView.self)
-
         NotificationCenter.default.addObserver(self, selector: #selector(positionsDidChange(notification:)), name: Notification.positionOthersChanged, object: nil)
     }
 
@@ -42,9 +40,5 @@ class BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoati
 
         // remove annotations that no longer exist
         mapView.removeAnnotations(unmatchedAnnotations)
-    }
-    
-    override func prepareAnnotationView(annotation: BikeAnnotation) -> BikeAnnoationView? {
-        mapView.dequeueReusableAnnotationView(ofType: BikeAnnoationView.self, with: annotation)
     }
 }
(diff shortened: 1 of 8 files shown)


CriticalMass/BikeAnnotationController.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @objc private func positionsDidChange(notification: Notification) {
+        guard let response = notification.object as? ApiResponse else { return }
+        display(locations: response.locations)
+    }
+
+    private func display(locations: [String: Location]) {
+        guard LocationManager.accessPermission == .authorized else {
+            return
```
mltbnz: maybe you can log a message here

CriticalMass/MKMapView+Register.swift
```diff
@@ -0,0 +1,34 @@
...
+import MapKit
+
+extension MKAnnotationView {
+    fileprivate class var reuseIdentifier: String {
+        return String(describing: self)
+    }
+}
+
+extension MKMapView {
+    func register<T: MKAnnotationView>(annotationViewType: T.Type) {
```
mltbnz: 👍

CriticalMass/UIView+Autolayout.swift
```diff
@@ -22,4 +22,15 @@ extension UIView {
             widthAnchor.constraint(equalToConstant: size.width)
         ])
     }
+    
+    func addLayoutsSameSizeAndOrigin(in view: UIView) {
+        translatesAutoresizingMaskIntoConstraints = false
+        
+        view.addConstraints([
+            NSLayoutConstraint(item: self, attribute: .width, relatedBy: .equal, toItem: view, attribute: .width, multiplier: 1, constant: 0),
```
mltbnz: I would opt in again for anchors here. Even Apple does not like its convenience method 😄

> Unlike the approach taken by the layout anchor API, you must specify a value for each parameter, even if it doesn’t affect the layout. The end result is a considerable amount of boilerplate code, which is usually harder to read. For example, the code in Listing 13-2 code is functionally identical to the code in Listing 13-1.

https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/AutolayoutPG/ProgrammaticallyCreatingConstraints.html

PR #198, opened 2019-11-22
Updated dependencies

lennet:
I run the update_dependencies script + Bumped the min version of SDWebImage to 5.3


lennet:
Good catch! Fixed that

PR #200, opened 2019-11-24
Prepare lokalize flow and add tooling

mltbnz:
Adding `BartyCrouch` and `SwiftGen` to the project to generate and organise Lokalize strings since it is potentially unsafe using `NSLocalizedString` directly which in some configurations is pointing to empty strings or non existing ones. Same goes for `UIImage(named: <ASSETNAME>)`

This PR will also use a new project file structure in which all resources live in a `Resources` folder which is why the Files changed is so huge. Makes this one hard to merge though.

### Before merging the PR


lennet:
Is SwiftGen required to build the project after this change or just needed to update translations & assets?

mltbnz:
No. Not mandatory. Like you said it is just to generate the code

mltbnz:
I am also not quite sure if that covers the workflow. Updating the strings can only be done by hand atm since the GitHub Integration does not do as I want (probably since I am not a manager of the cm repo) and updating them via script and the Crowdin Api does not work because I would need to expose the api key in some way. So it is not a complete round two like I would have wanted. Still an improvement though

PR #201, opened 2019-12-01
Swiftlint rules update

mltbnz:
This PR will bump up the `.swiftversion` to 5.1 so bye bye `return` for single line functions etc.
It will also disable the `trailingComma` rule since it is not helping the codebase IMHO.

**FROM**
```swift
return [1,2,3,4,]
```

**TO**
```swift
return [1,2,3,4]
```

### Before merging the PR


lennet:
I didn't know about the benefits of having trailing comas http://www.trailingcommas.com but I don't have any strong opinions about having or not having them

mltbnz:
Didn’t know the website.
I am fine with both. Personally I don’t see the benefit of cleaner diffs though there is a point to it. I don’t want this chance to be personal preference and if you feel like this is taking away something rather than adding to the codebase I will remove it

fbernutz:
> Didn’t know the website.
> I am fine with both. Personally I don’t see the benefit of cleaner diffs though there is a point to it. I don’t want this chance to be personal preference and if you feel like this is taking away something rather than adding to the codebase I will remove it

It's fine for me to disable the rule.

Issue #96, opened 2019-05-15
Friends


lennet:
Open Questions/Tasks
- [x] Verification (Signature?)
- [x] Verification caching
- [x] Maps UI
- [x] Settings UI
- [x] Storage
- [ ] documentation

lennet:
related to https://github.com/criticalmaps/criticalmaps-android/issues/119

zutrinken:
There is a UI suggestion in the iOS design files (last two screen on the right): <img src="https://raw.githubusercontent.com/criticalmaps/criticalmaps-design/master/ios/critical-maps-app-preview.png" />

But that's not based on the latest technical constraints we discussed. So for a UI concept we first need a description on how the feature can be technically implemented.

lennet:
Yeah we still don't really decided what we'll do. I'll start implementing the settings ui based on the code I already have

lennet:
@zutrinken How about adding an empty state view/ message for users that don't have friends connected yet. Maybe a small explanation 
<img width="584" alt="Bildschirmfoto 2019-09-23 um 22 06 58" src="https://user-images.githubusercontent.com/7677738/65458698-7992f400-de4e-11e9-87bb-7cc8e90f6c06.png">

maxxx777:
Would love to see this feature in the app. 
What is the current state of the implementation? 
If I can help with the feature, please let me know.

lennet:
We have a feature branch (`friends`) for that  with a fully working version of that feature however it's not ready for production yet. 


There are some things to be done 
1. The Map overlays need to be updated to support displaying names 
<img width="504" alt="Bildschirmfoto 2019-10-06 um 17 52 08" src="https://user-images.githubusercontent.com/7677738/66271734-059b1780-e862-11e9-82b2-89feea3843b1.png">
The full design specs are available in our design repo https://github.com/criticalmaps/criticalmaps-design

2. We are not 100% sure how we should do the the friends recognition with our current API. We are currently renewing the users id every 24 hours for privacy reasons which makes it hard to recognise people (which is also the purpose of that :D). There are some possible options 
- Don't renew the id for some days/ weeks which makes people auto expiring and also everyone could track users over multiple days as long as their using the friends feature
- Sign the id with an key and share the key for verifying the signature with your friends so that your Id can change and friends need to verify the signature (This method is currently implemented in the feature branch). This is comes with some overhead as a users needs to try verifying the signature for every id on the map as long as they haven't found their friends. 
[...]

PR #204, opened 2019-12-04
Add Friends Feature (hidden behind feature toggle)

fbernutz:
This adds the friends feature.

- [x] after merging #169 in `friends` branch -> merge changes in this branch
- [x] Check which files are correct
- [x] Localize new Strings
- [x] Refactoring (Map related code)
- [x] Fix SPM Build in CI
- [x] Use Feature Gate to hide Friends Feature (still under development)


CriticalMass/MapViewController.swift
```diff
@@ -186,15 +229,42 @@ extension MapViewController: MKMapViewDelegate {
...
+                                             reuseIdentifier: BikeAnnoationView.reuseIdentifier)
+                }
+            } else {
+                view = BikeAnnoationView(annotation: userAnnotation,
+                                         reuseIdentifier: BikeAnnoationView.reuseIdentifier)
+            }
         }
-
-        return mapView.dequeueReusableAnnotationView(ofType: matchingController.annotationViewType, with: annotation)
+        return view
```
lennet: Maybe we should revert that change and fix the FriendAnnotation in a future PR

CriticalMass/MapViewController.swift
```diff
@@ -135,7 +138,38 @@ class MapViewController: UIViewController {
         mapView.showsUserLocation = true
     }
 
-    // GPS Disabled Overlay
+    private func display(locations: [String: Location]) {
```
lennet: This method now lives in BikeAnnotationController

lennet: We can remove it an fix the friend related code in a Future PR

CriticalMass/AppDataStore.swift
```diff
@@ -0,0 +1,149 @@
...
+
+import CoreData
+import UIKit
+
+public class AppDataStore: DataStore {
+    private var userDefaults: UserDefaults
+
+    public init(userDefaults: UserDefaults = .standard) {
+        self.userDefaults = userDefaults
+        loadFriends()
```
lennet: We should do a FeatureCheck here otherwise we would create an SQLLite Database in production. We shouldn't do that because there might be spec changes before we ship the feature which might require database migrations

CriticalMass/AppController.swift
```diff
@@ -111,10 +111,25 @@ class AppController {
...
+        SettingsViewController(themeController: themeController, dataStore: dataStore, idProvider: idProvider)
     }
 
     private func loadInitialData() {
         requestManager.getData()
     }
+
+    public func handle(url: URL) -> Bool {
+        do {
+            let followURLObject = try FollowURLObject.decode(from: url.absoluteString)
```
lennet: We also need a Feature check here

CriticalMass/MapViewController.swift
fbernutz: I'll add a new issue for this 👍

fbernutz: https://github.com/criticalmaps/criticalmaps-ios/issues/217

CriticalMass/MapViewController.swift
fbernutz: https://github.com/criticalmaps/criticalmaps-ios/issues/217

fbernutz:
@lennet do you think we can close this issue https://github.com/criticalmaps/criticalmaps-ios/issues/96 after merging this PR? I think it would be better to open new issues with smaller tasks when there is something missing.

lennet:
Yes

PR #206, opened 2019-12-08
Add Feature enum to activate or deactivate features

### LINKED ISSUE #203: Feature flagging
We have some difficulties currently to maintain the Friends feature branch. It would be nice to have a small light weight figure flagging component that would allow us to merge the Friends feature in to the master branch without being available in a production environment

fbernutz:
This adds a simple `Feature` enum to activate or deactivate features. Fixes #203.

## Usage
To write code which is dependent of whether a feature is active or not, use the following code: 

```swift 
if Feature.friends.isActive {
    // show friends feature
} else {
    // hide friends feature
}
```

To activate the feature, the developer has to set the flag to true in `Feature.swift`. 

```swift
private var currentState: [Feature: Bool] = [
    .friends: true
]
```

To permanently activate a feature for example in tests, use this: 

```swift 
var feature = Feature.friends
feature.isActive = true
```

## Info

With this implementation it is not possible to activate a feature during runtime with user input (but certainly could be added).


lennet:
I really like the separation between the Feature enum and the storage and state. I'm wondering if we could expose the state outside the file and inject the state into files that needs them. That would make automated testing easier to build

CriticalMass/Feature.swift
```diff
@@ -0,0 +1,23 @@
...
+private var currentState: [Feature: Bool] = [
+    .friends: false
+]
+
+enum Feature {
+    case friends
+
+    var isActive: Bool {
+        set { currentState[self] = newValue }
+        get { currentState[self] ?? false }
```
mltbnz: To me this feels overcomplicated. Why not use a simple `Bool`?
Thing that is bugging me is the nil-coalescing operator for asking the dict

fbernutz: Sorry, I don't quite understand what you mean. 
You need something to store the current state for the Feature in, this is `currentState` as private dictionary, so you save the state of each feature in it, if it is activated or not. To get the value if a feature is active or not, you can use `Feature.friends.isActive` which returns a `Bool` and it reads out the value from the dictionary. If the dictionary does not contain the Feature the default for the value is `false`. 
Of course you could make `currentState` public and read out and write in there directly from everywhere, but I prefer the API `Feature.isActive` over `currentState[Feature.friends]`. 

I would be interested, which idea you had in mind. Can you share a code snippet?

mltbnz: My idea is to go with something simple like
```swift
enum FeatureFlags {
    static let friends = true
}
```

The proposed solution is fine though. Just want to avoid `??` if possible

mltbnz: And the API callside is nicer with an isActive property. 

But maybe that could be a computed property which directly maps the enum cases to bools?

mltbnz:
```swift
var isActice: Bool {
    switch self {
        case .friends: true
    }
}
```

fbernutz: That are nice ideas, they are even simpler! 
A disadvantage with these is though, that you can't change the feature toggle specific for tests. I imagine a snapshot test (for example) which should test a snapshot with activated feature toggle, so when removing the feature toggle the test is still successful. 
Therefore you could add this as test preparation and test with the "new" feature: 

```swift
override func setUp() {
    super.setUp()

    var feature = Feature.friends
    feature.isActive = true
}
```

But maybe we never need these kind of features and we should add these things, when we really need them. So I think I'd go with the `static let` version. What would you think?

lennet: We already have tests that would require a Feature check (e.g. AppDataStoreTests). 
The Static API looks super nice but makes testing super and I don't like having a global static state that controls the logic. I would prefer passing the state  around with dependency injection like
```swift
featureState.isEnabled(.friends)
```

For tests we could inject a different featureState with the  feature enabled

fbernutz: ok I'll see, will try that

lennet: But that's just my opinion if you both prefer a nicer API that is static that is also fine with me as long as we can somehow enable or disable features for tests

fbernutz: Sorry, I don't have much time lately. 🙈If it's fine for you both to start with my current solution, I would be happy if we merge it like that. Feel free to refactor it later, if you want to.

lennet: No worries! Let’s do that

Issue #194, opened 2019-11-10
Next ride info


zutrinken:
Show an info box for the next ride nearby, based on the [criticalmass.in](https://criticalmass.in) API.
* The most nearby event is shown by a special marker.
* Tab on info box centers map on event location.
* Info box and marker disappear 30min after the start of current event.
* Info box and marker for the next event appear 8h after the start of previous event.

<table><tr><th>Light</th><th>Dark</th></tr><tr><td>
<img src ="https://user-images.githubusercontent.com/888679/68546887-d2592480-03db-11ea-8d4c-2f8deea8adbf.png" /></td><td><img src ="https://user-images.githubusercontent.com/888679/68546888-d2592480-03db-11ea-8311-091c9e47e0bc.png" /></td></tr></table>

mltbnz:
Maybe the feature could also get its own tab where we could mirror some of the functionality from httpss://criticalmass.in/
Maybe on the Map the App can Show the next Mass closest to your coordinate. Therefore the api would need an endpoint to do so. 
The tab could also have the option to add new made to the Database

maltehuebner:
I’m already working on it in https://github.com/calderacc/criticalmass-in/pull/567

The endpoint will allow you to create complex queries for dates and cities and of cource you may limit the results to a bounding box or a radius around a provided coordinate.

maltehuebner:
Okay, it took me some more days, but most of the api is now covered by tests and it looks like it’s working now: https://criticalmass.in/api/doc

Please have a look at it and tell me if it’s useful for you.

lennet:
I think we might start with a simple first version as presented by @zutrinken and iterate instead of starting a big project that’s hard to finish. Is the design spec final enough to create smaller issues for the feature ?

PR #208, opened 2019-12-14
Added MapInfoView

lennet:
This PR adds MapInfoView, a component which we can use for presenting error messages to the use and providing more information on the map (#194 )
It can be configured with a message and two styles

Alert:
Bildschirmfoto 2019-12-14 um 10 47 02

Info:
Bildschirmfoto 2019-12-14 um 10 49 32

Dynamic Type, VoiceOver are supported 

Todos:
- [x] Darkmode
- [x] Snapshottests


CriticalMass/MapInfoView.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    private var configuration: Configuration?
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        layer.cornerRadius = 18
+        layer.shadowColor = UIColor.black.cgColor
+        layer.shadowRadius = 4
+        layer.shadowOpacity = 0.2
+        layer.shadowOffset = CGSize(width: 0, height: 2)
```
mltbnz: maybe this can be a setupLayer func like
```swift
private func setup(_ layer: CALayer) {
    layer.cornerRadius = 18
    layer.shadowColor = UIColor.black.cgColor
    layer.shadowRadius = 4
    layer.shadowOpacity = 0.2
    layer.shadowOffset = CGSize(width: 0, height: 2)
}
```

and then 

```swift
setup(layer)
```

CriticalMass/MapViewController.swift
```diff
@@ -83,12 +85,43 @@ class MapViewController: UIViewController {
         gpsDisabledOverlayView.set(title: String.mapLayerInfoTitle, message: String.mapLayerInfo)
         gpsDisabledOverlayView.addButtonTarget(self, action: #selector(didTapGPSDisabledOverlayButton))
         view.addSubview(gpsDisabledOverlayView)
+        gpsDisabledOverlayView.addLayoutsSameSizeAndOrigin(in: view)
+        updateGPSDisabledOverlayVisibility()
+    }
+
+    public func presentMapInfo(with configuration: MapInfoView.Configuration) {
+        dismissMapInfo()
```
mltbnz: Maybe use a FeatureFlag here if it is not fully implemented

CriticalMass/MapViewController.swift
```diff
@@ -41,8 +41,8 @@ class MapViewController: UIViewController {
 
     private var mapView = MKMapView(frame: .zero)
 
-    private let gpsDisabledOverlayView: BlurryOverlayView = {
-        let view = BlurryOverlayView.fromNib()
+    private let gpsDisabledOverlayView: BlurryFullscreenOverlayView = {
```
mltbnz: This would be a perfect candiate for the PropertyWrapper that we had earilier :)

CriticalMass/MapViewController.swift
```diff
@@ -83,12 +85,43 @@ class MapViewController: UIViewController {
...
+            view.widthAnchor.constraint(lessThanOrEqualToConstant: 400),
+            view.centerXAnchor.constraint(equalTo: self.view.centerXAnchor),
         ])
 
-        updateGPSDisabledOverlayVisibility()
+        UIAccessibility.post(notification: .layoutChanged, argument: view)
+    }
+
+    public func dismissMapInfo() {
+        view.subviews
```
mltbnz: Couldn't you use the `gpsDisabledOverlayView ` property directly here without going through the view hierarchy?

CriticalMass/MapViewController.swift
```diff
@@ -83,12 +85,43 @@ class MapViewController: UIViewController {
...
+    }
+
+    public func presentMapInfo(with configuration: MapInfoView.Configuration) {
+        dismissMapInfo()
+
+        let view = MapInfoView.fromNib()
+        view.configure(with: configuration)
+        self.view.addSubview(view)
+
+        view.translatesAutoresizingMaskIntoConstraints = false
```
mltbnz: would move `translatesAutoresizingMaskIntoConstraints` under to the `view` creation

CriticalMass/MapViewController.swift
```diff
@@ -61,6 +61,8 @@ class MapViewController: UIViewController {
             .forEach(mapView.register)
 
         setNeedsStatusBarAppearanceUpdate()
+
+        presentMapInfo(with: MapInfoView.Configuration(title: "Foo", style: .info))
```
mltbnz: ?

CriticalMass/MapViewController.swift
lennet: What was the property wrapper again? I remember that there was a cool one but can't find it anymore

CriticalMass/MapInfoView.swift
lennet: Apparently,  the navigation overlay uses the same shadow and corner radius parameters. I moved the code to an extension

CriticalMass/MapViewController.swift
lennet: Ups🙄

CriticalMass/MapViewController.swift
lennet: I don't think we need a FeatureFlag here since nobody is calling them method for now (as long as I don't commit my debug calls :D )

CriticalMass/MapViewController.swift
lennet: Yes, you are right, the current solution isn't great . I'm thinking about introducing a MapOverlayViewController in a future PR. to unify the overlay specific code without making the MapViewController more complicated. Especially to make sure that only one overlay is visible at a time etc.

CriticalMass/MapViewController.swift
mltbnz: It’s gone. Made us loose the translates... (unsure how the property is called) property. Maybe I’ll add it again and adopt it to other properties too

mltbnz: ... in another PR

PR #210, opened 2019-12-17
Improved Accessibility for NavigationOverlayViewController

lennet:
Adding the tabbar traits enables some VoiceOver improvements out of the box such as quicker navigation and VoiceOver automatically announces "Button label 1 out of N" instead of of just "Button label" which is also the default behaviour of a UITabBarController.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changelog for Critical Maps iOS
 
 - Replace send text button in ChatInputView with an icon button
 - Accessability support for chat input textview
+- VoiceOver improvements for the Navigation TabBar
 
 ## [3.6.0] - 2019-11-28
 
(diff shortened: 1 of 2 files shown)


mltbnz:
I remember that the phone as the first thing was reading the statusbar to me. Is that something we can influence with a accessibility property? (Not related to the PR I guess)

lennet:
> Mh, I tried it with both, simulator (iOS 13.3) and on my own device (iOS 13.3), and the 4 buttons in the "tab bar" only have the trait "button" and voice over is not announcing the "x of n" ...

I was confused as well by the inspectors output. I'll file a radar for that

lennet:
> I remember that the phone as the first thing was reading the statusbar to me. Is that something we can influence with a accessibility property? (Not related to the PR I guess)
> 
> I would think that the first thing should be navigation

We could influence that but its super risky  to manually move the focus because I might be confusing if the device automatically moves to focus to a Point what a User wouldn't expect. The. default focus navigation is always from the top left to the bottom right (for left to right languages) which is correct here. Adding the tabbar trait also gives VoiceOver the ability to quickly navigate to the navigation without going through all map related elements

PR #215, opened 2019-12-21
Moved core logic into Swift Packages

lennet:
This experimental PR moves the core logic of the app into two Swift Packages:
1. CriticalMapsFoundation
2. CriticalMapsKit

Bildschirmfoto 2019-12-20 um 17 42 30

This could enable user to have a clean separation between the core logic and the rest of the app and enables us to share the code with other targets in the future like a watch app or a today extension for the next ride info feature (#194). 

On the other side, we are losing our git history and there will be many merge conflicts 🙈


fbernutz:
I think it's a good idea to move the core logic into own packages so we can use them in the iOS App and watch App.

There are some thoughts about that: 
- When I think about adding a new Feature, it would not be 100% clear to me, which logic should be implemented in which package. Especially `CriticalMapsKit` and `CriticalMapsUtilsKit` sound way too similar. 
- At the moment there is no Xcode project where we can run the iOS app... ?

lennet:
How about CriticalMapsFoundation? 

The old Xcode project can still be used to build the App

fbernutz:
I think `Foundation` is way better than `UtilsKit`, but it's really hard for me to understand the difference between the content of the different packages. Maybe you could add a description to both packages with examples, so it gets more clear what the content of the package is and should be. For example, why are the Colors currently in the `CriticalMass App` and not in `CriticalMapsKit`? I think a Watch App should use the same colors like the app... 

Just an idea, maybe one package `CriticalMapsKit` would be enough for the moment and when we need more separate modules we split the package again? 

In general, I'm fine with this structure, I would just search for a file in both packages. But when adding a new feature I would be lost where things have to go. 🙈

mltbnz:
I would also struggle to find the right kit for adding new code. I think a union of the mentioned is good for a start.

mltbnz:
Not sure about the merge conflicts which would mean all other tasks or PRs should maybe rest for a while until this can be merged but that is fine for me.

lennet:
I'll abandon this PR for now. Because it's too hard to keep it in sync with the master branch.

We should restart the initiative when we actually need it

PR #221, opened 2020-01-13
Upload UITest screenshots to test results

lennet:
I reenabled Screenshots for UITests and added a step to upload them as CI Artifacts, so we can download them for every pull request build.

Bildschirmfoto 2020-01-13 um 22 35 03

We could use artefacts for other cool stuff as well like uploading the code coverage or snapshots if the snapshot test failed

(I hope you didn't get as many emails about failing test as I did 🙈)


mltbnz:
Cool.
Maybe clean up the commit history a bit 😂

.github/workflows/main.yml
```diff
@@ -34,12 +34,21 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v1
 
+    - name: Edit Snapfile
+      run: echo "devices([\"iPhone 8\"]);languages([\"en-US\"])" > fastlane/Snapfile
```
fbernutz: Is this really needed? I would love to use it without this, because now, we have to define the language and test device on two different places (`Fastfile` and here)...

.github/workflows/main.yml
```diff
@@ -34,12 +34,21 @@ jobs:
...
+
     - name: Install Gem Dependencies
       run: bundle install
 
-    - name: Run UnitTests
```
fbernutz: Ups 😇 Thanks for the fix

fbernutz:
> Cool.
> Maybe clean up the commit history a bit 😂

💯 
Or simply use the merge feature `squash and merge` ;)

.github/workflows/main.yml
lennet: The problem is that fastlane fails to run a snap command if the snap file contains a device that is not available on the CI (I this case the Se) even if the the devices isn‘t required in the snap command. Not sure if we can delete the snap file completely.

I’ll try that if that doesn’t work I’ll add a comment to explain why this line is needed

fbernutz: What do you think about moving all the code from the Snapfile to the Fastfile? So we have all the devices and languages defined in the Fastfile.

lennet: looks way better now!

PR #222, opened 2020-01-16
Add userDefaults to extension

mltbnz:
What should have been a hands-on for PropertyWrappers turned out to be just a small refactoring for the use of `UserDefaults` in our app. Since they were kind of all around I wanted to centralize them and remove some boilerplate code. Using the Wrapper turned out to be not that handy with the current structure so I ended up streamlining the use of UserDefaults a bit

### Before merging the PR

### DIFF
--- a/CriticalMass/AppDataStore.swift
+++ b/CriticalMass/AppDataStore.swift
@@ -22,7 +22,7 @@ public class AppDataStore: DataStore {
     private var lastKnownResponse: ApiResponse?
 
     public var userName: String {
-        get { userDefaults.username }
+        get { userDefaults.username ?? UIDevice.current.name }
         set { userDefaults.username = newValue }
     }
 
--- a/CriticalMass/AppDataStore.swift
+++ b/CriticalMass/AppDataStore.swift
@@ -9,7 +9,7 @@ import CoreData
 import UIKit
 
 public class AppDataStore: DataStore {
-    private var userDefaults: UserDefaults
+    private let userDefaults: UserDefaults
 
     public init(userDefaults: UserDefaults = .standard) {
         self.userDefaults = userDefaults
@@ -22,15 +22,8 @@ public class AppDataStore: DataStore {
     private var lastKnownResponse: ApiResponse?
 
     public var userName: String {
-        set {
-            if !newValue.isEmpty {
-                userDefaults.set(newValue, forKey: #function)
-            }
-        }
-
-        get {
-            userDefaults.string(forKey: #function) ?? UIDevice.current.name
-        }
+        get { userDefaults.username }
+        set { userDefaults.username = newValue }
     }
 
     public func update(with response: ApiResponse) {
--- a/CriticalMass/Date+Additions.swift
+++ b/CriticalMass/Date+Additions.swift
@@ -4,6 +4,20 @@
 import Foundation
 
 public extension Date {
-    static var yesterday: Date { Date(timeInterval: -86400, since: .now) }
+    static var yesterday: Date {
+        let dateComponents = DateComponents(day: -1)
+        return Calendar.current.date(byAdding: dateComponents, to: .now)!
+    }
+
     static var now: Date { Date() }
+
+    /// Get a component representation of todays Date as Int.
+    /// - Parameter keyPath:
+    /// - Returns: DateComponent representation as Int. Returns 0 when component is not available
+    static func getCurrent(_ keyPath: KeyPath<DateComponents, Int?>) -> Int {
+        let calendar = Calendar.current
+        let components = calendar.dateComponents([.year, .month, .day], from: .now)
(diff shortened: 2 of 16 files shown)


CriticalMass/RatingHelper.swift
```diff
@@ -17,17 +17,19 @@ public protocol RatingRequest {
 }
 
 public class RatingHelper {
-    public var daysUntilPrompt = 1
-    public var usesUntilPrompt = 5
+    private let daysUntilPrompt = 1
```
mltbnz: maybe I am wrong here but I saw no point in having these as public vars

mltbnz:
GitHub CI is really fast!

CriticalMass/AppDataStore.swift
```diff
@@ -22,15 +22,8 @@ public class AppDataStore: DataStore {
...
-            if !newValue.isEmpty {
-                userDefaults.set(newValue, forKey: #function)
-            }
-        }
-
-        get {
-            userDefaults.string(forKey: #function) ?? UIDevice.current.name
-        }
+        get { userDefaults.username ?? UIDevice.current.name }
+        set { userDefaults.username = newValue }
```
lennet: We need to check if the new name isn't an empty string. But we can also move that check to the UI

CriticalMass/Date+Additions.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  CriticalMaps
+
+import Foundation
+
+public extension Date {
+    static var yesterday: Date { Date(timeInterval: -86400, since: .now) }
```
lennet: I'm always scared of adding seconds to a Date because a Date might contain leap seconds.. 
Better use Calendar for that https://developer.apple.com/documentation/foundation/calendar/2293676-date

CriticalMass/Preferences.swift
```diff
@@ -6,40 +6,20 @@
 //
 
 import Foundation
+import UIKit
```
lennet: why do wee need an UIKit import now?

CriticalMass/Theme.swift
```diff
@@ -8,7 +8,7 @@
 
 import UIKit
 
-enum Theme: Int {
+enum Theme: String {
```
lennet: This change means that users would loose their current preference right?

CriticalMassTests/RatingHelperTests.swift
```diff
@@ -19,83 +19,70 @@ class MockRatingRequest: RatingRequest {
 
 class RatingHelperTests: XCTestCase {
```
lennet: Love how simple the tests are now!

CriticalMass/Theme.swift
mltbnz: Oh. Not sure 🤔 
But I think so yes. Since it wants a string from the defaults it should return nil and then be a light theme by default on <iOS 13

CriticalMass/Date+Additions.swift
mltbnz: Ah. So you mean this could end up in not being yesterday. Haven’t thought of that. But it is just wrapping what was in the tests before. But might be a good idea to rework to Calendar

CriticalMass/Preferences.swift
mltbnz: Can go I think. 
Was there because of the defaults

CriticalMass/AppDataStore.swift
mltbnz: Think it is better located there since we would want to react to it, right?

mltbnz: Just had a look and I am not sure where to actually change it in the UI. `FriendSettingsTableViewCell`?

CriticalMass/Date+Additions.swift
mltbnz: Updated to Calendar and added some UnitTests

CriticalMass/AppDataStore.swift
lennet: We can just ignore the input if it's empty and not pass it to the DataStore and show the old input again. That is also the behaviour of the current implementation

CriticalMass/Theme.swift
lennet: We should probably revert that change to prevent data loss

mltbnz: Since it is just a theme change and on older OS I would merge it since it makes the API clearer imo

lennet: Yes it makes it clearer but in my opinion the benefits are not strong enough to accept loosing the preference but I’m also okay with merging it

CriticalMass/AppDataStore.swift
mltbnz: So where to do that? I honestly don’t know where to adjust

PR #223, opened 2020-01-19
Converted ErrorHandler into a Protocol

lennet:
I converted ErrorHandler into a protocol to make it easier to inject different error handler for different use cases and prototyped a MapOverlayErrorHandler (behind a Feature flag)

ErroHandler

I'm not super happy with my implementation (especially AppController becomes hard to understand with all the inner dependencies between the lazy variables) would appreciate some feedback!


mltbnz:
Hard to change the implementation I think since we mostly want to inject the properties I would say

lennet:
I'll try to refactor that in a future PR

PR #224, opened 2020-01-20
Add next ride map info view and annotation

mltbnz:
## Additions
* CMMarkerAnnotationView (iOS 11+) and CMAnnotationView (iOS 10)
* `NextRideManager` to load, filter and sort upcoming events
* `CoordinateObfuscator` to obfuscate userCoordinate

## Updates
* move MapInfoView logic to separate controller

## Additions
* animate present and dismiss of MapInfoView
* add swipe up gesture to dismiss

## Screenshot
Screenshot 2020-03-03 at 13 43 27

### Before merging the PR


CriticalMass/MapInfoViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+        ) {
+            self.infoViewContainerTopConstraint.constant = Constants.infoBarVisible
+            self.view.layoutIfNeeded()
+        }
+        animator.startAnimation()
+
+        UIAccessibility.post(notification: .layoutChanged, argument: view)
+    }
+
+    public func dismissMapInfo(_: Bool = false) {
```
mltbnz: I hate SwiftFormat for this...

lennet: +1! We can disable the rule (unusedArguments)

mltbnz: !!! would be very happy if we do

mltbnz:
So the BE gives me an object like 
```json
{
        "id": 7728,
        "slug": null,
        "title": "Critical Mass Berlin 31.01.2020",
        "description": null,
        "dateTime": 1580497200,
        "location": "Mariannenplatz",
        "latitude": 52.505355445221,
        "longitude": 13.418254852295,
        "estimatedParticipants": null,
        "estimatedDistance": null,
        "estimatedDuration": null
    }
```

and for the MapInfoView title I simply used the objects title but I think that works.

mltbnz:
What does not work well is that the every internal user of the app will get next events in germany so I think we need to somehow add a filter for them...

Screenshot 2020-01-20 at 17 35 36


Or I think the BE would send an empty collection back (which is also not handled atm)

lennet:
I would could filter results based on the distance. We are more flexible If we do that locally on the device

mltbnz:
@zutrinken Could you provide me a monochrome version of the logo in black and white? I used the magic wand of Preview to do that on a croped screenshot of the Website. Not ideal I would say 😸

mltbnz:
> We could filter results based on the distance. We are more flexible If we do that locally on the device

I added a `NextRideManager` to filter too far away rides ✨

CriticalMass/MapViewController.swift
```diff
@@ -221,6 +229,24 @@ extension MapViewController: MKMapViewDelegate {
...
+            } else {
+                if let cmAnnotationView = mapView.dequeueReusableAnnotationView(withIdentifier: CMAnnotationView.reuseIdentifier) {
+                    cmAnnotationView.annotation = annotation
+                    return cmAnnotationView
+                } else {
+                    return CMAnnotationView(annotation: annotation, reuseIdentifier: CMAnnotationView.reuseIdentifier)
+                }
+            }
+        }
+
```
lennet: We should reuse the AnnotationController & AnnotationView infrastructure for that

CriticalMass/MapViewController.swift
```diff
@@ -183,18 +176,25 @@ class MapViewController: UIViewController {
...
+                guard let nextRide = ride else {
+                    Logger.log(.debug, log: .map, "Expected nextRide to present MapInfoView")
+                    return
+                }
+                onMain { [weak self] in
+                    self?.mapInfoViewController.presentMapInfo(title: nextRide.title, style: .info)
+                    let cmAnnotation = CriticalMassAnnotation(ride: nextRide)
+                    self?.cmAnnotation = cmAnnotation
+                    self?.mapView.addAnnotation(cmAnnotation)
+                }
```
lennet: That should be in an AnnotationController

CriticalMass/MapViewController.swift
```diff
@@ -183,18 +176,25 @@ class MapViewController: UIViewController {
...
+        focusOnCoordinate(CLLocationCoordinate2D(location))
 
-        // TODO: Replace test implemenation with controller based
-        guard Feature.events.isActive else { return }
         let coordinate = CLLocationCoordinate2D(
             latitude: location[keyPath: \Location.latitude],
             longitude: location[keyPath: \Location.longitude]
         )
-        nextRideHandler.getNextRide(around: coordinate) { result in
+        nextRideManager.getNextRide(around: coordinate) { result in
```
lennet: We should think about anonymising the data to hide the true location.

This can be done by adding a random offset in a given range or clip decimals

CriticalMass/MapViewController.swift
mltbnz: I will take a look at it

CriticalMass/MapViewController.swift
mltbnz: What exactly? The API call?

CriticalMass/MapViewController.swift
mltbnz: Ah. Good point. I can add that

CriticalMass/MapViewController.swift
mltbnz: Might be that my Generics knowledge does not reach far enough but I think Swifts Type inference is not able to add elements to `[AnnotationController]` if the other controller use different types then the `BikeAnnotation` and `BikeAnnotationView`

mltbnz: I updated the `AnnotationController` since it did not needed the Generic constraints IMO

CriticalMass/MapViewController.swift
mltbnz: I am adding a random double in the Range of `(-0.009 ... 0.009)` to lat and lon

CriticalMass/MapViewController.swift
mltbnz: I think now I know.
I checked the AnnotationController and now I think I know what its for 👍

CriticalMass/AppController.swift
```diff
@@ -53,15 +53,13 @@ class AppController {
         TwitterManager(networkLayer: networkOperator, request: TwitterRequest())
     }()
 
-    private lazy var mapOverlayErrorHandler: ErrorHandler = {
```
mltbnz: I removed that after the rebase since it was not used and I added a `MapInfoViewController` @lennet

mltbnz: I also did not get why this has to be injected and the presentInfo function was public?

CriticalMass/MapViewController.swift
lennet: Oh right :/ 

The only solution that came to my mind is adding a Protocol around it which looks like that

```swift
protocol AnnotationControllable {
    var annotationType: IdentifiableAnnnotation.Type { get }
    var annotationViewType: MKAnnotationView.Type { get }
}

class AnnotationController<T: IdentifiableAnnnotation, K: MKAnnotationView> {
    private let _annotationType = T.self
    private let _annotationViewType = K.self

    [...]
}

extension AnnotationController: AnnotationControllable {
    var annotationType: IdentifiableAnnnotation.Type {
        _annotationType
    }

    var annotationViewType: MKAnnotationView.Type {
        _annotationViewType
    }
}
```

And then use AnnotationControllable as a type. The generic constraint is needed to make sure at compile time that MKMapView+Register never fails. We can't safely make that assumption if the annotationType and annotationViewType is mutable

CriticalMass/MapViewController.swift
```diff
@@ -181,35 +178,61 @@ class MapViewController: UIViewController {
...
-        focusOnLocation(location: location)
-
-        // TODO: Replace test implemenation with controller based
-        guard Feature.events.isActive else { return }
-        let coordinate = CLLocationCoordinate2D(
-            latitude: location[keyPath: \Location.latitude],
-            longitude: location[keyPath: \Location.longitude]
-        )
-        nextRideHandler.getNextRide(around: coordinate) { result in
+    private func getNextRide(_ coordinate: CLLocationCoordinate2D) {
```
lennet: We should move that ride logic to the AnnotationController. The idea behind AnnotationController is that it contains or controls all the logic that is need to display annotations

CriticalMass/AppController.swift
lennet: That was the problem I was describing in my PR. There is a circular dependency and by removing it. the code compiles but the error won't be presented

CriticalMass/MapViewController.swift
mltbnz: But since it is a MKAnnotation and MKAnnotationView it will always register them and then we have to ask if the annotation is of type `Something` anyways, right?

CriticalMass/AppController.swift
mltbnz: Ah :) It feels a little weird. Should we remove it again for now and tackle that later?

CriticalMass/MapViewController.swift
mltbnz: But this call also triggers the MapInfoView

mltbnz: I didn't really get the AnnotationController tbh. I moved the Annotation related logic into a AnnotationController but also having the API call which has other effects into that and then reaching out again does not make sense to me

mltbnz:
Also these damn Snapshot tests keep failing although I exchanged the images? Do I need to clean that up somewhere?

CriticalMass/MapViewController.swift
lennet: Yes it will also work, it will just break if you accidentally use the wrong initialiser for a given AttentionController. Another benefit of the Protocol is that all the code is in a single place and you don't need to override every initialiser

CriticalMass/MapViewController.swift
lennet: I would like to keep the MapViewController as stupid as possible because I'm worried that it might become a "Godclass" at some point. 
The idea is that the MapViewController is just just contains logic to communicate between the AnnotationController and the Map. The Content related logic (Bikes, Events and Friends) should be in specific AnnotationController which makes it also easier to test.

To workaround that specific need, you could inject the MapInfoViewController into the AnnotationController. But I'm also okay with keeping it as it is for now and doing that in a Future PR

CriticalMass/MapInfoViewController.swift
```diff
@@ -0,0 +1,103 @@
+//
+//  CriticalMaps
+
+import CoreLocation
+import UIKit
+
+// https://material.io/design/motion/speed.html#easing
+extension UICubicTimingParameters {
```
fbernutz: Should this be a private extension which is only used in this file or do we possibly need it somewhere else? When we possibly need it somewhere else, I would prefer to put this in an own file.

CriticalMass/MapViewController.swift
```diff
@@ -63,12 +78,28 @@ class MapViewController: UIViewController {
         configureTileRenderer()
         configureMapView()
         condfigureGPSDisabledOverlayView()
+        registerAnnoationViews()
+        setupMapInfoViewController()
+
+        setNeedsStatusBarAppearanceUpdate()
+    }
 
+    private func registerAnnoationViews() {
```
fbernutz: There is a typo in here, I think it should be `registerAnnotationViews()` 🙂

CriticalMassTests/NextRideManagerTests.swift
```diff
@@ -0,0 +1,174 @@
+//
+//  CriticalMapsTests
+
+import CoreLocation
+@testable import CriticalMaps
+import XCTest
+
+class NextRideManagerTests: XCTestCase {
+    var sut: NextRideManager!
```
fbernutz: What does `sut` mean? I would prefer `nextRideManager` as variable name.

CriticalMass/MapViewController.swift
mltbnz: I agree with the benefits of a protocol

CriticalMassTests/NextRideManagerTests.swift
mltbnz: it usually refers to `system under test` but we have a very loose naming pattern for that

CriticalMass/MapViewController.swift
mltbnz: Like that idea. If a future PR is okay then I'll go with that option 🏗

mltbnz: Having references to MapViewController properties in these controllers makes it harder to reason about these I think since they have shared state. Maybe having one instead of one for each content type. But not making the Map a god makes sense for sure

CriticalMass/MapViewController.swift
mltbnz: But there is only one init for the controller and the types are constraint to the specific controller. So I do not see where it should break. But I can rework to the protocol solution

CriticalMassTests/NextRideManagerTests.swift
fbernutz: Oh, didn’t know that! Now it makes a lot more sense 🙈

lennet:
The feature is currently missing logic around when to show & his the Annotation (specs for that are in the issue description). This PR is already quite big, I would suggest reading the Feature flag and adding that in a separate PR

mltbnz:
I added a check to remove the annotation in the controller. 👍 
Alos readded the flag for now since the feature might need some adjustments here and there

mltbnz:
Gonna reopen when ready

stephanlindauer:
btw: do we have some attribution to @maltehuebner for using `criticalmass.in` data?

mltbnz:
Don’t think we do
Where would you put it?

stephanlindauer:
Screenshot_20200219-133550
we have this "about" section in android where we would probably add it when we start using it.

mltbnz:
> we have this "about" section in android where we would probably add it when we start using it.

We could add an about section in the settings tab I guess or me could mention it in the README

lennet:
> btw: do we have some attribution to @maltehuebner for using `criticalmass.in` data?

CC @zutrinken

zutrinken:
An attribution view for `Used libraries` is nice to have. At the end of settings under info. Maybe an autogenerated list with all libraries. If you click on one it shows you its licence.

The personal attribution is already handled by the `Contributors` list on the [website](https://www.criticalmaps.net/info#Contributors).

For [criticalmass.in](https://criticalmass.in) another card as the yellow GitHub one would be nice. Could need some input here like "_Missing a mass?- Submit your bike activism on criticalmass.in and promote it to other bike enthusiasts_".

mltbnz:
We have the license acknowledgments in the apps settings under system settings

CriticalMass/NextRideManager.swift
```diff
@@ -40,7 +40,9 @@ class NextRideManager {
     }
 
     private func getUpcomingRide(_ rides: [Ride]) -> Ride? {
-        rides.first { $0.dateTime > .now }
+        rides
+            .sorted(by: \.dateTime)
+            .first { $0.dateTime > .now }
```
lennet: A small improvement could be using a lazy colelction here 

```suggestion
        rides
            .lazy      
            .sorted(by: \.dateTime)
            .first { $0.dateTime > .now }
```

mltbnz:
Updated with localised date and time:
Screenshot 2020-03-03 at 13 43 27

mltbnz:
Closed in favor of a new PR with the same code but cleaner commit history (i. rebase was not an option)
See #253

PR #227, opened 2020-01-21
add Hound CI config

mltbnz:
To have some automated linting on PRs that are created I would like to add https://houndci.com/ to the repo. This PR is the setup to that.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,0 +1,2 @@
+swift:
+  config_file: .swiftlint.yml
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,31 @@
+disabled_rules:
+  - variable_name
+  - nesting
+  - function_parameter_count
+opt_in_rules:
+  - control_statement
+  - empty_count
+  - trailing_newline
+  - colon
+  - comma
+included:
+  - CriticalMaps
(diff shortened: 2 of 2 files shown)


lennet:
What are the benefits of introducing a second linter? Shouldn’t we automatically enforce the rules similar to what we are currently doing with swiftformat?

mltbnz:
https://github.com/nicklockwood/SwiftFormat#linting
It is not actually a real linter and also we never enforce the use of it in the project setup afaik.
The Benefit of having the Linter on PRs IMO is that it definitely gets enforced and also SwiftLint does a better job on enforcing a coding style and is super configurable

lennet:
Yeah it's not the same but the Idea. I worked with SwiftLint before and it was super frustrating because it was always complaining :D. I always prefer tools that automatically fix my code. SwiftLint also has a mode that does that. We should just focus on one tool because we might end up with conflicting rule sets

mltbnz:
Ok. I guess it is personal preference then

mltbnz:
Think that Hound or any other real linting solution would benefit the repo more than automatic code formatting with SwiftFormat. SwiftLint needs some adjustment and rules that everyone is convenient to follow but it then is super powerful

Issue #217, opened 2019-12-24
Friends: Fix Friend Annotation


fbernutz:
There are two patches which help adding a `FriendAnnotation`. But this should be cleaned up and be fixed, because it's not working correctly. 

The Friends Feature is hidden behind a Feature Toggle. To use this Feature, go to `Feature.swift` and change this:

```swift
private var currentState: [Feature: Bool] = [
    .friends: true
]
```

<details>
<summary>patch 1</summary>

```
diff --git a/CriticalMass/BikeAnnotationController.swift b/CriticalMass/BikeAnnotationController.swift
index 1abfe70..b5435d3 100644
--- a/CriticalMass/BikeAnnotationController.swift
+++ b/CriticalMass/BikeAnnotationController.swift
@@ -29,6 +29,13 @@ class BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoati
         var unmatchedAnnotations: [MKAnnotation] = []
         // update existing annotations
         mapView.annotations.compactMap { $0 as? BikeAnnotation }.forEach { annotation in
+            if friendsVerificationController.isFriend(id: annotation.identifier) {
+                annotation.type = .friend
+                annotation.friend = friendsVerificationController.friend(for: annotation.identifier)
+            } else {
+                annotation.type = .user
+                annotation.friend = nil
+            }
             if let location = unmatchedLocations[annotation.identifier] {
                 annotation.location = location
                 unmatchedLocations.removeValue(forKey: annotation.identifier)
@@ -36,7 +43,7 @@ class BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoati
                 unmatchedAnnotations.append(annotation)
             }
         }
-        let annotations = unmatchedLocations.map { BikeAnnotation(location: $0.value, identifier: $0.key) }
+        let annotations = unmatchedLocations.map { IdentifiableAnnnotation(location: $0.value, identifier: $0.key) }
         mapView.addAnnotations(annotations)
 
         // remove annotations that no longer exist
```

</details>

<details>
<summary>patch 2</summary>

```
diff --git a/CriticalMass/MapViewController.swift b/CriticalMass/MapViewController.swift
index ee459f5..66d375f 100644
--- a/CriticalMass/MapViewController.swift
+++ b/CriticalMass/MapViewController.swift
@@ -196,15 +196,42 @@ extension MapViewController: MKMapViewDelegate {
     // MARK: MKMapViewDelegate
 
[...]

PR #228, opened 2020-01-22
Fixed friends annotations on Map

lennet:
This PR fixes FriendsAnnotation #217

The AnnotationViewController isn't instantiated in MapViewController yet because of some problems with generics we are currently discussing in #224


CriticalMass/AnnotationController.swift
```diff
@@ -19,4 +19,25 @@ class AnnotationController<T: IdentifiableAnnnotation, K: MKAnnotationView> {
...
+        // update existing annotations
+        mapView.annotations.compactMap { $0 as? T }.forEach { annotation in
+            if let location = unmatchedLocations[annotation.identifier] {
+                annotation.location = location
+                unmatchedLocations.removeValue(forKey: annotation.identifier)
+            } else {
+                unmatchedAnnotations.append(annotation)
+            }
+        }
+        let annotations = unmatchedLocations.map { T(location: $0.value, identifier: $0.key) }
```
mltbnz: Would every AnnotationController need to implement `IdentifiableAnnnotation` which is a subclass of `MKPointAnnotation`? Since `T: IdentifiableAnnnotation`

lennet: Yes we need identifiable annotations to be able to perform diffing operations on displayed annotations

lennet: This specific line might need to change because there is currently no way to inject data

lennet: I added a mechanism to inject data into Annotations if needed

CriticalMass/BikeAnnotationView.swift
```diff
@@ -7,7 +7,15 @@
 
 import MapKit
 
-class BikeAnnoationView: MKAnnotationView {
+protocol AnnotationView: MKAnnotationView {
+    associatedtype T = Any
```
mltbnz: Do you need to pin Any to an associatedType and then in an implementation say `T = Friend`?

CriticalMass/BikeAnnotationView.swift
```diff
@@ -26,16 +34,12 @@ class BikeAnnoationView: MKAnnotationView {
     }
 }
 
-class FriendAnnotationView: MKAnnotationView {
+class FriendAnnotationView: MKAnnotationView, AnnotationView {
+    typealias T = Friend
```
mltbnz: I think it should not be named `T` since this is not a Generic and it mixes things up

CriticalMass/FriendAnnotationController.swift
```diff
@@ -0,0 +1,55 @@
...
+    private func display(locations: [String: Location]) {
+        guard Feature.friends.isActive else {
+            return
+        }
+        guard LocationManager.accessPermission == .authorized else {
+            Logger.log(.info, log: .map, "Bike annotations cannot be displayed because no GPS Access permission granted", parameter: LocationManager.accessPermission.rawValue)
+            return
+        }
+        let filtredLocations = locations.filter { friendsVerificationController.isFriend(id: $0.key) }
+        let allTuples: [(identifier: String, location: Location, object: Friend)] = filtredLocations.compactMap {
```
mltbnz: maybe define a typealias for `(identifier: String, location: Location, object: Friend)`

CriticalMass/AnnotationController.swift
```diff
@@ -8,7 +8,7 @@
 
 import MapKit
 
-class AnnotationController<T: IdentifiableAnnnotation, K: MKAnnotationView> {
+class AnnotationController<T: IdentifiableAnnnotation<J>, K: AnnotationView, J> {
```
mltbnz: I feel like the Generics are just just solving the friends use case but maybe I do not see the bigger picture. I could of course check other annotations too but might not be needed for them

CriticalMass/BikeAnnotationView.swift
lennet: The Any is a default value in case the AnnotationView doesn't need an Object.  Swift currently doesn't support optional Associated type so this is a workaround for that.

CriticalMass/AnnotationController.swift
lennet: The Generics can be used for any AnnotationController that needs to provide an Object to their annotation views in a type safe way. That could also be used for events

CriticalMass/FriendAnnotationController.swift
lennet: I have problems finding a name that makes sense :D

CriticalMass/AnnotationController.swift
mltbnz: On a second look I see how this is supposed to work.
Though I can not see the type safety in this `BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoationView, Any>` and I do not see this working in an Array like in `MapViewController`

What I like is that there is one function to update annotations that I can override for my needs 👍 

What I do not like about is that: 
* `T` needs to be an IdentifiableAnnotation since that is MKPointAnnotation which is a concrete implementation  and not the protocol `MKAnnotation` 
* Makes this class hard to read and wrap heads around with a lot of wrapping and generic Types
* Uses `Location` type and not `CLLocation` which is preferred IMO (But I guess that is due to the LocationManager sending out Location types)
* What I meant in the beginning: It feels like the class was designed to fulfil the needs of a specific AnnotationType. So maybe it would make sense to have an empty `updateAnnotations` func since subclassing of the controller is necessary anyways

CriticalMass/BikeAnnotationController.swift
```diff
@@ -25,21 +36,12 @@ class BikeAnnotationController: AnnotationController<BikeAnnotation, BikeAnnoati
...
-        var unmatchedAnnotations: [MKAnnotation] = []
-        // update existing annotations
-        mapView.annotations.compactMap { $0 as? BikeAnnotation }.forEach { annotation in
-            if let location = unmatchedLocations[annotation.identifier] {
-                annotation.location = location
-                unmatchedLocations.removeValue(forKey: annotation.identifier)
-            } else {
-                unmatchedAnnotations.append(annotation)
-            }
+        var filtredLocations = locations
```
mltbnz: Typo

CriticalMass/IdentifiableAnnnotation.swift
```diff
@@ -7,25 +7,20 @@
...
 import MapKit
 
-class IdentifiableAnnnotation: MKPointAnnotation {
-    enum UserType: Int, Decodable {
-        case friend
-        case user
-    }
-
-    var type: UserType = .user
+class IdentifiableAnnnotation<T>: MKPointAnnotation {
```
mltbnz: Could this be just MKAnnotation instead to make it more versatile for other annotation types for example

CriticalMass/MapViewController.swift
```diff
@@ -225,7 +225,12 @@ extension MapViewController: MKMapViewDelegate {
             return nil
         }
 
-        return mapView.dequeueReusableAnnotationView(ofType: matchingController.annotationViewType, with: annotation)
+        let castedAnnotation = unsafeBitCast(annotation, to: matchingController.annotationType)
```
mltbnz: Using this speaks against toe safety to me.

Apple docs:
Warning
Calling this function breaks the guarantees of the Swift type system; use with extreme care.

CriticalMass/BikeAnnotationView.swift
```diff
@@ -72,9 +76,12 @@ class FriendAnnotationView: MKAnnotationView {
 
         friendView = FriendView(frame: .init(x: 0, y: 0, width: 109, height: 23))
         friendView!.backgroundColor = .clear
-        friendView!.name = friend?.name ?? ""
         addSubview(friendView!)
     }
+
+    func set(object: Friend?) {
```
mltbnz: Should be `func set(object: Object?)` IMO when using typealias

CriticalMass/MapViewController.swift
lennet: Yes, I'm aware of that. However, the generic constraint that is used to define `matchingController.annotationViewType` makes sure that this is "safe" to use here (famous last words :D)

CriticalMass/IdentifiableAnnnotation.swift
lennet: Not sure if that's possible. MKAnnotation is just a protocol and not a class. When I use MKAnnotation the annotation don't seem to update anymore

CriticalMass/AnnotationController.swift
lennet: Oh I totally forgot about the array 🙈 Trying to find a solution for that

lennet: Sorry! you were totally right with your scepticism with using Generics here. Didn't see all the issues coming 🤦‍♂️ Not sure if it's  still worth it moving forward with this PR.. 

I removed the Array and replaced it with some properties witch introduced some code duplication in MapViewController. The code with all it generic Types is not nice to read. Not sure if the benefits of separation of concerns are strong enough to keep it

CriticalMass/IdentifiableAnnnotation.swift
mltbnz: 🤔 I would assume it would still work since Confirming to the protocol would require to add the coordinate property but it is there already, right?

mltbnz: Maybe you need to define the coordinate property as `@objc dynamic` like in https://developer.apple.com/documentation/mapkit/mapkit_annotations/annotating_a_map_with_custom_data

mltbnz:
So what do you think about the implementation of #224. It is a bit verbose but could work as a starting point to ist on it from there. Since it is working as an array type and do the job as before I would propose to use it and work on it from there.
I actually haven’t tested it with the simulation mode though

lennet:
> So what do you think about the implementation of #224. It is a bit verbose but could work as a starting point to ist on it from there. Since it is working as an array type and do the job as before I would propose to use it and work on it from there.
> I actually haven’t tested it with the simulation mode though

Yes, lets youse your solution as a Starting point. I'll put this PR on hold until yours is merged

mltbnz:
Wouldn’t it be possible to use type erasure  for wrapping the controller? Something like `AnyAnnotationController` that wraps the controller to use it in an array?

Used a lot in combine land 😄

lennet:
> Wouldn’t it be possible to use type erasure for wrapping the controller? Something like `AnyAnnotationController` that wraps the controller to use it in an array?
> 
> Used a lot in combine land 😄

Yes, that works, but also requires removing the generic constraints. I'll probably start over that work next week

mltbnz:
Since leo can't work on this anymore I think I'll close the PR

PR #241, opened 2020-02-19
Update BikeAnnotationView appearance

mltbnz:
Will update the BikeAnnotationView to the Zeplin designs.

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot 2020-02-19 at 11 03 32 Screenshot 2020-02-19 at 11 03 38 | Screenshot 2020-02-19 at 10 45 36 Screenshot 2020-02-19 at 10 45 45 |

## Checklist


CriticalMass/BikeAnnotationView.swift
```diff
@@ -21,8 +38,12 @@ class BikeAnnoationView: MKAnnotationView {
     }
 
     private func commonInit() {
-        image = UIImage(named: "Bike")
         canShowCallout = false
+        backgroundColor = .clear
+
+        frame = Constants.shapeRect
```
lennet: I’m wondering if we should slightly increase the size based on dynamic type preferences.

mltbnz: I like that idea

mltbnz: But I think that would mean I would have to iterate through all the annotations and alter their views... 😩 Or is there a nicer way? Maps does not have that for its annotations afaik so maybe I could just increase the initial size 🤷‍♂️

lennet: BikeAnnotationView could listen to https://developer.apple.com/documentation/uikit/uicontentsizecategory/1622948-didchangenotification and update itself. But we can also create an issue for that and follow up

mltbnz: I did a quick test where BikeAnnotationView reacts to traitcollectiondidchange and although I was able to update the view bounds and map the preferredcontentsize to a new size it did not trigger a redraw of the shape. Seems like I forgot something

mltbnz: But I will create an issue for that and merge

PR #247, opened 2020-02-27
Fixed delay for displaying Annotations on Map

lennet:
I was experiencing a delay for displaying Annotations but couldn't really find the issue but I accidentally found it while debugging #228 .

Because didSet won't get called on Init we are not setting the Coordinate Attribute which is needed to display the Annotation on a Map. So we were just showing Annotations that were shown before which als results in an empty Map for (minimum) 12 seconds after launching the App.

We should probably send out a new version with this fix

### DIFF
--- a/CriticalMass/IdentifiableAnnnotation.swift
+++ b/CriticalMass/IdentifiableAnnnotation.swift
@@ -19,7 +19,7 @@ class IdentifiableAnnnotation: MKPointAnnotation {
 
     var location: Location {
         didSet {
-            coordinate = CLLocationCoordinate2D(latitude: location.latitude, longitude: location.longitude)
+            coordinate = CLLocationCoordinate2D(location)
         }
     }
 
@@ -27,5 +27,6 @@ class IdentifiableAnnnotation: MKPointAnnotation {
         self.identifier = identifier
         self.location = location
         super.init()
+        coordinate = CLLocationCoordinate2D(location)
     }
 }
--- a/CriticalMassTests/IdentifiableAnnotationTests.swift
+++ b/CriticalMassTests/IdentifiableAnnotationTests.swift
@@ -1,0 +1,27 @@
+//
+//  IdentifiableAnnotationTests.swift
+//  CriticalMapsTests
+//
+//  Created by Leonard Thomas on 27.02.20.
+//  Copyright © 2020 Pokus Labs. All rights reserved.
+//
+
+@testable import CriticalMaps
+import MapKit
+import XCTest
+
+extension CLLocationCoordinate2D: Equatable {
+    public static func == (lhs: CLLocationCoordinate2D, rhs: CLLocationCoordinate2D) -> Bool {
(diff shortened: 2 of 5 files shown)


CriticalMass/IdentifiableAnnnotation.swift
```diff
@@ -19,13 +19,14 @@ class IdentifiableAnnnotation: MKPointAnnotation {
...
-            coordinate = CLLocationCoordinate2D(latitude: location.latitude, longitude: location.longitude)
+            coordinate = CLLocationCoordinate2D(location)
         }
     }
 
     init(location: Location, identifier: String) {
         self.identifier = identifier
         self.location = location
         super.init()
+        coordinate = CLLocationCoordinate2D(location)
```
mltbnz: You could set location in a defer statement which would trigger didSet. That way coordinated would get set only from one place afai can see in review mode on my phone.

CriticalMassTests/IdentifiableAnnotationTests.swift
```diff
@@ -0,0 +1,27 @@
...
+import MapKit
+import XCTest
+
+extension CLLocationCoordinate2D: Equatable {
+    public static func == (lhs: CLLocationCoordinate2D, rhs: CLLocationCoordinate2D) -> Bool {
+        lhs.latitude == rhs.latitude && lhs.longitude == rhs.longitude
+    }
+}
+
+class IdentifiableAnnotationTests: XCTestCase {
```
mltbnz: I like that bugs now have a test as well 😍 ☑️

CriticalMass/IdentifiableAnnnotation.swift
fbernutz: Interesting, didn't know `defer` could help in this case (I found an example here: https://stackoverflow.com/a/33979852/6716961). But this can be only used with Optionals, which we could do but I don't know if it's worth it. 

It would then look like this: 
```swift
var location: Location? {
        didSet {
            guard let location = location else { return }
            coordinate = CLLocationCoordinate2D(location)
        }
    }

    init(location: Location, identifier: String) {
        self.identifier = identifier
        defer {
            self.location = location
        }
        super.init()
    }
```

mltbnz: Right. With `Location` as the type for this it would need to be optional. I am wondering though why we pass that type around anyways since it is just the coordinate that is needed. When that would be CLCoordinate2D the initial state could be `kCLLocationCoordinate2DInvalid` and then use defer. 
But I guess that would need a rebuild of the location manager which is sending out the` Location ` type which I think is a flaw. I am working on a rewrite of the manager btw. as a first step to move away from notification based data flow which is often used

mltbnz: +++ Breaking +++
Just read that the defer usage in the init is considered a bug by the swift core team 😅 So better not use it

PR #248, opened 2020-02-29
Fix Dark Mode Color Behaviour

### LINKED ISSUE #175: The action indicator is hardly visible in Dark Mode

fbernutz:
This fixes some dark mode color behaviour for:
- action indicator (RulesViewController) 
- action indicator (SettingsViewController) -> Fixes #175
- separator (SettingsViewController) 

We need some of the mock classes in the Snapshot Tests now, so I moved the classes from `TestHelper.swift` in `CriticalMassTests` to separate files in `CriticalMassMocksForTests`. 

### Before merging the PR

- [x] If applicable, did you add a before / after screenshot of the feature? -> see SnapshotTests


CriticalMass/ThemeController.swift
```diff
@@ -69,7 +69,6 @@ class ThemeController {
         // Custom Views
         SettingsFooterView.appearance().versionTextColor = theme.titleTextColor
         SettingsFooterView.appearance().buildTextColor = theme.titleTextColor
-        SettingsGithubTableViewCellTableViewCell.appearance().arrowTintColor = .settingsOpenSourceForeground
```
fbernutz: This line made problems after adding the SnapshotTest and it is not really needed here, because the `arrowTintColor` is not set dependent on the theme.

fbernutz:
I have no clue why the snapshot tests fail on CI, running the tests in Xcode and with fastlane (locally) succeeds :( any idea?

CriticalMass/RulesViewController.swift
```diff
@@ -31,17 +31,50 @@ enum Rule: String, CaseIterable {
 
 class RulesViewController: UITableViewController {
     private let rules = Rule.allCases
+    private let themeController: ThemeController!
```
mltbnz: Is the explicit IUO needed since it is passed in the init?

fbernutz: IUO? I need to get access to the currentTheme in `updateThemeIfNeeded()`, so I adapted the implementation from `MapViewController` and pass the themeController in the init.

fbernutz:
> The PR title confused me when I saw 23 changes but looks good.
> Could you elaborate why the Mock classes are needed?

I updated the PR title, I hope it fits better now. Most of the changes are from adding SnapshotTests and from moving the mock classes. 
The mock classes were already implemented, I just moved them to `CriticalMassMocksForTests` and split them into separate files because I now needed some of them in the Snapshot Tests. So, there's nothing new from me in the mock classes.

CriticalMass/RulesViewController.swift
mltbnz: Sorry. I mean implicitly unwrapped optional
``` swift 
private let themeController: ThemeController
```
This should work, right?

fbernutz: Ah of course! Yes, the implicitly unwrapped optional is not needed 👍 Will remove it

PR #250, opened 2020-03-01
update pull_request_template.md

mltbnz:
**👆This will go since nobody deleted the damn thing (except core team ❤️)**

Updates the PR Template for a better PR experience :)
Kudos to https://github.com/kickstarter/ios-oss since it is more or less theirs

👇 **THIS BECOMES MORE EXPLICIT**
### Before merging the PR

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -26,9 +26,3 @@ Screenshots, external resources?
 - [ ] Tap targets use minimum of 44x44 pts dimensions
 - [ ] Works with VoiceOver
 - [ ] Supports Dynamic Type 
-
-## ⏰ TODO
-
-- [ ] Got planned refactors to come
-- [ ] Will add more tests
-- [ ] Need feedback on a design
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -15,6 +15,7 @@ Some background context on why the change is needed.
 ## 👀 See
 
 Screenshots, external resources?
+- [ ] Added snapshot tests (using iPhone 8 Simulator)
 
 | Before 🐛 | After 🦋 |
 | --- | --- |
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,33 @@
-## Thank you for your contribution! 
+<!-- This template is **just a guide**, delete any and all parts which you don't need! -->
 
-Please add one or two sentences _why_ this PR was made, and optionally what it adds/changes…
+## 📝 Docs
 
-## Checklist
-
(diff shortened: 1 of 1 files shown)


mltbnz:
Think GitHub still has issues ...
We could add an action for first time contributors to say thanks 🙏

.github/pull_request_template.md
```diff
@@ -1,11 +1,33 @@
...
-- [ ] Did you update the [CHANGELOG.md](https://github.com/criticalmaps/criticalmaps-ios/blob/master/CHANGELOG.md)?
+A description of the change.
+
+## 🤔 Why
+
+Some background context on why the change is needed.
+
+## 👀 See
+
+Screenshots, external resources?
```
fbernutz: Do we want to add that (if possible) a Snapshot Test should be added and that is has to be recorded with an iPhone 8 (13.3)?

.github/pull_request_template.md
```diff
@@ -1,11 +1,33 @@
...
+
+- [ ] Tap targets use minimum of 44x44 pts dimensions
+- [ ] Works with VoiceOver
+- [ ] Supports Dynamic Type 
+
+## ⏰ TODO
+
+- [ ] Got planned refactors to come
+- [ ] Will add more tests
+- [ ] Need feedback on a design
```
fbernutz: Should you check the field when there are open refactors or should it be unchecked because the refactors are still open? These three points would confuse me a bit.. 🤔

fbernutz:
> Think GitHub still has issues ...

There is a problem on the checkout step, could you try merge the latest master (or something like that)? Seems weird that it fails on this step.

> We could add an action for first time contributors to say thanks 🙏

I love the idea to say thanks 😍

.github/pull_request_template.md
mltbnz: I would check them if for example there is something I want to improve on an implementation but it would bloat the PR or if I want to add more tests

mltbnz: As a way to say not done yet I think

.github/pull_request_template.md
mltbnz: Good idea. Got confused a couple of times with failed tests

mltbnz:
Here is the action. 
https://github.com/marketplace/actions/first-interaction

.github/pull_request_template.md
fbernutz: In my opinion, I would leave these points out. I think, it wouldn't be very often that we would actually have to use them.

mltbnz: I removed them for now 👍

PR #252, opened 2020-03-03
Proposal: Reuse URLCodable for APIRequestDefining queryItems

lennet:
We could reuse the logic from URLCodable to define the queryItems in APIRequestDefining using Codable Objects

*Pros*
- Typesafety
- Less code 

*Cons*
- Using String as a default associatedType for QueryDefiningDataType to make the compiler happy in case we don't have query Items for a Request doesn't look super nice


CriticalMass/URLCodable.swift
```diff
@@ -46,28 +46,34 @@ extension URLCodable {
         return try self.init(scheme: urlComponents.scheme, host: urlComponents.host, path: urlComponents.path, queryObject: codableObject)
     }
 
-    private func encode<T: Encodable>(codable: T) throws -> [URLQueryItem] {
+    private static func decode<T: Codable>(type: T.Type, from items: [URLQueryItem]) throws -> T {
+        let dict = items.reduce(into: [String: String]()) { result, item in
+            result[item.name] = item.value
+        }
+        let data = try JSONSerialization.data(withJSONObject: dict, options: [])
+        return try JSONDecoder().decode(type, from: data)
```
mltbnz: could be
```suggestion
        return try data.decoded()
```

CriticalMass/URLCodable.swift
```diff
@@ -46,28 +46,34 @@ extension URLCodable {
...
+    }
+}
+
+extension URLQueryItem {
+    static func encode<T: Encodable>(codable: T) throws -> [URLQueryItem] {
         let jsonData = try JSONEncoder().encode(codable)
```
mltbnz: could be:
```suggestion
        let jsonData = try codable.encoded()
```

CriticalMass/APIRequest.swift
```diff
@@ -5,16 +5,24 @@ enum APIRequestBuildError: Error {
...
-    var queryItems: [URLQueryItem]? { get }
+    var queryItem: QueryDefiningDataType? { get }
     var requiresBackgroundTask: Bool { get }
     func makeRequest() throws -> URLRequest
     func parseResponse(data: Data) throws -> ResponseDataType
 }
 
+extension APIRequestDefining {
+    var queryItem: QueryDefiningDataType? {
+        nil
```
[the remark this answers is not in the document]
mltbnz: So no need to explicitly return that as a default

CriticalMass/URLCodable.swift
```diff
@@ -46,28 +46,34 @@ extension URLCodable {
         return try self.init(scheme: urlComponents.scheme, host: urlComponents.host, path: urlComponents.path, queryObject: codableObject)
     }
 
-    private func encode<T: Encodable>(codable: T) throws -> [URLQueryItem] {
+    private static func decode<T: Codable>(type: T.Type, from items: [URLQueryItem]) throws -> T {
```
mltbnz: Maybe the function should be renamed since it does a lot more than just to decode

CriticalMass/URLCodable.swift
```diff
@@ -46,28 +46,34 @@ extension URLCodable {
...
+        let dict = items.reduce(into: [String: String]()) { result, item in
+            result[item.name] = item.value
+        }
+        let data = try JSONSerialization.data(withJSONObject: dict, options: [])
+        return try JSONDecoder().decode(type, from: data)
+    }
+}
+
+extension URLQueryItem {
+    static func encode<T: Encodable>(codable: T) throws -> [URLQueryItem] {
```
mltbnz: Could also be renamed IMO

CriticalMass/URLCodable.swift
lennet: Nice that helper method also enables us to remove the QueryDefiningDataType workaround 🎉

CriticalMass/APIRequest.swift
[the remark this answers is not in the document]
lennet: Extensions only support computed properties so we have to explicitly return that. 

Having a default implementation in an extension in this case has two benefits
1. APIRequestDefining implementations that don't need a queryItem don't need to implement it
2. Implementations that need to define can override it with a specific type instead of Encodable

CriticalMass/URLCodable.swift
lennet: In my opinion the name makes sense because it's a known pattern from the Codable protocol. Example
```swift
let followURLObject = try FollowURLObject.decode(from: url.absoluteString)
```

But I'm open for suggestions

CriticalMass/APIRequest.swift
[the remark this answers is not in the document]
mltbnz: But this does not need to have a default implementation in the extension since it is `nil` already

CriticalMass/URLCodable.swift
mltbnz: Maybe then pull
```swift
return try dict.lazy.map { (key, value) -> URLQueryItem in
            switch value {
            case let stringValue as String:
                return URLQueryItem(name: key, value: stringValue)
            case let stringConvertible as CustomStringConvertible:
                return URLQueryItem(name: key, value: String(describing: stringConvertible))
            default:
                // More types are currently not supported and should be added if needed
                throw URLCodableError.encodingFailed
            }
        }.sorted { a, b in
            // The sorting is only required to make testing easier
            a.name > b.name
        }
```
out?

CriticalMass/APIRequest.swift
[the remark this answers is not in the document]
lennet: Yes, but without that default implementation, each APIRequestDefining implementation has to implement `var queryItem: String?` which I'm trying to avoid. Especially since assigning a String doesn't really work because String is just a placeholder type. 
Can also remove that and add the implementations.

CriticalMass/URLCodable.swift
lennet: I simplified the method a bit. Is it better now?

CriticalMass/APIRequest.swift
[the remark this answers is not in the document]
mltbnz: True. Missed that.
Maybe a builder instead of the make request function in the protocol would make things clearer. But not for this pr

CriticalMass/URLCodable.swift
```diff
@@ -46,28 +46,29 @@ extension URLCodable {
...
-    }
-
-    private static func decode<T: Codable>(type: T.Type, from items: [URLQueryItem]) throws -> T {
-        let dict = items.reduce(into: [String: String]()) { result, item in
-            result[item.name] = item.value
-        }
-        let data = try JSONSerialization.data(withJSONObject: dict, options: [])
-        return try JSONDecoder().decode(type, from: data)
+        // The sorting is only required to make testing easier
+        .sorted(by: \.name)
```
fbernutz: good idea 👍

CriticalMass/APIRequest.swift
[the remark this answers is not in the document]
lennet: Yes, also though about experimenting with compositing multiple protocols to eliminate optionals in a future PR

PR #254, opened 2020-03-03
Update locationmanager

mltbnz:
## 📝 Docs

## 📲 What

A description of the change.

## 🤔 Why

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

## ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type

### DIFF
--- a/CriticalMass/LocationManager.swift
+++ b/CriticalMass/LocationManager.swift
@@ -7,14 +7,37 @@
 
 import CoreLocation
 
-class LocationManager: NSObject, CLLocationManagerDelegate, LocationProvider {
+public enum GeolocationRequestError: Error {
+    case denied
+    case error
+}
+
+public enum GeolocationAuthResult {
+    /// user have permission on geolocation
+    case success
+    /// user haven't permission on geolocation
+    case denied
+    /// user doesn't give permission on his geolocation
+    case failure
+    /// system dialog reqeusts user permission at this moment
+    case requesting
+}
+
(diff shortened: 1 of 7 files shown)


CriticalMass/LocationManager.swift
```diff
@@ -24,75 +47,117 @@ class LocationManager: NSObject, CLLocationManagerDelegate, LocationProvider {
...
+    private let locationManager: CLLocationManager
+
+    init(
+        manager: CLLocationManager = CLLocationManager(),
+        locationProperties: LocationProperties = LocationProperties()
+    ) {
+        locationManager = manager
+        super.init()
+        configureLocationManager(with: locationProperties)
+        startTrackingLocation()
```
lennet: Not sure if we still need. to call stratTrackingLocation on Init since the getCurrentLocation method is doing. that anyway

CriticalMass/MapViewController.swift
```diff
@@ -49,25 +60,40 @@ class MapViewController: UIViewController {
...
+                break
+            }
+        }
+    }
+
+    deinit {
+        NotificationCenter.default.removeObserver(self, name: .themeDidChange, object: nil)
+        NotificationCenter.default.removeObserver(self, name: .focusLocation, object: nil)
+        NotificationCenter.default.removeObserver(self, name: .observationModeChanged, object: nil)
+    }
```
lennet: This is not needed anymore. Especially since we never deinit MapViewController

PR #256, opened 2020-03-05
Converted testing to xctestplan

lennet:
## 📲 What

I converted testing go xctestplan and enabled parallel and random test order

## 🤔 Why

Testplans allow multiple configuration which might be useful in the future. Parallelizable should slightly improve the total test duration.


fbernutz:
Is this testplan already in use? When I inspect the schemes they all use the different tests like before (and the button with "Convert to use test plans") is visible. 

But I haven't used test plans before and am not sure how these should work.

CriticalMaps.xcodeproj/CriticalMaps.xctestplan
```diff
@@ -0,0 +1,47 @@
+{
+  "configurations" : [
+    {
+      "id" : "3B90BF13-D62B-459B-B9C6-5162A3DEF527",
+      "name" : "Configuration 1",
```
mltbnz: Maybe name it default since this is the one that is executed mostly

lennet:
I had to disable parallel execution because it kept crashing on CI

CriticalMapsSnapshotTests/CriticalMapsSnapshotTests.xctestplan
```diff
@@ -0,0 +1,28 @@
+{
+  "configurations" : [
+    {
+      "id" : "9675AA6D-95DA-4EBD-985A-7278586C91C6",
+      "name" : "Configuration 1",
```
fbernutz: I think this should be called `Default` as well. 

```suggestion
      "name" : "Default",
```

CriticalMassTests/CriticalMaps.xctestplan
```diff
@@ -0,0 +1,44 @@
...
+      "options" : {
+
+      }
+    },
+    {
+      "id" : "FF294556-426B-40CD-A500-D4998EA2E3B9",
+      "name" : "Configuration 3",
+      "options" : {
+
+      }
```
fbernutz: `Configuration 2` and `Configuration 3` have no different setting than `Default` (as I can see). The unit tests are currently run three times with these settings. I think these two additional configurations can be deleted. 

```suggestion
```

lennet:
Not sure what's going on with `testNetworkSatisfiedStatusUpdate`. Seems to fail sometimes with testplans

fbernutz:
> Not sure what's going on with `testNetworkSatisfiedStatusUpdate`. Seems to fail sometimes with testplans

It failed locally for me sometimes too... let's see

lennet:
Maybe the random test order makes it more likely to fail 🤔

fbernutz:
Do we want to merge it anyway? 
I think when the CI breaks too often, it would be better to disable the test until the issue is fixed. So in my opinion, this can be merged 👍

PR #261, opened 2020-03-07
Add UITest for adding a Friend from URL

### LINKED ISSUE #164: Friends Feature - UITests
It would be nice to have UITests for testing the full flow of adding a friend through opening a link (example link: criticalmaps:follow?token=TestKey&name=TestName) and deleting it afterwards in the friends view.

All the work friends feature can be found in the Friends feature branch

fbernutz:
## 📲 What
 
Adds a UITest for adding a Friend from URL. This fixes #164. 

## 🤔 Why

A test is useful for this and it actually found an issue with the format of the URL. 
To open Critical Maps app from Safari the format of the URL has to look like this `criticalmaps://{scheme}` instead of `criticalmaps:{scheme}`.


CriticalMass/SettingsSection.swift
```diff
@@ -23,11 +23,13 @@ enum Section: CaseIterable {
         var title: String?
         var subtitle: String?
         var action: Action
+        let accessibilityIdentifier: String
```
lennet: Minor: can we make them all var or all let?

fbernutz: Good idea, they shouldn't change after initialization, so I decided to make them all `let`s :)

PR #264, opened 2020-03-09
Animate CriticalMass Annotation when added to the map

mltbnz:
## 📲 What

Set CriticalMass marker annotation `animatesWhenAdded` property true

## 🤔 Why

Adds a nice touch to it when the annotation is added to the map in the visible area


lennet:
Can you add a small gif that demos the effect?

mltbnz:
> Can you add a small gif that demos the effect?

Sure 🙂
demo
_recorded with slow animations on_

mltbnz:
I just realized that someone should update the coordinate in criticalmass.in. Little far of

PR #265, opened 2020-03-09
Improved Code Coverage + Bug fixes

lennet:
## 📲 What

I added some tests and simplified some code to improve the test coverage of the project. While writing tests, I found and fixed a small bug in FormatDisplay when the difference between now and a Tweets created_date is less that 1 minute. I also removed the static date variables in DateFormatter that could end up in outdated time stamps for tweets if the app was open for a while.

## 🤔 Why

giphy-2

### DIFF
--- a/CriticalMass/FormatDisplay.swift
+++ b/CriticalMass/FormatDisplay.swift
@@ -34,7 +34,7 @@ enum FormatDisplay {
 
     static func dateString(for tweet: Tweet) -> String? {
         let components = Calendar.current.dateComponents(
-            [.minute, .hour, .day, .month],
+            [.second, .minute, .hour, .day, .month],
             from: tweet.created_at, to: currentDate
         ).dateComponentFromBiggestComponent
 
--- a/CriticalMass/FormatDisplay.swift
+++ b/CriticalMass/FormatDisplay.swift
@@ -29,11 +29,7 @@ private extension DateComponents {
 }
 
 enum FormatDisplay {
-    // inject current date to be able to keep the same time difference between tweet date and current date
-    static var currentDate = Date()
-    static var calendar = Calendar.current
-
-    static func dateString(for tweet: Tweet) -> String? {
+    static func dateString(for tweet: Tweet, currentDate: Date = Date(), calendar: Calendar = .current) -> String? {
         let components = calendar.dateComponents(
             [.second, .minute, .hour, .day, .month],
             from: tweet.created_at, to: currentDate
@@ -46,9 +42,10 @@ enum FormatDisplay {
         return formatter.string(from: components)
     }
 
-    static func hoursAndMinutesDateString(from message: ChatMessage) -> String {
+    static func hoursAndMinutesDateString(from message: ChatMessage, calendar: Calendar = .current) -> String {
         let date = Date(timeIntervalSince1970: message.timestamp)
         let formatter = DateFormatter()
+        formatter.calendar = calendar
         formatter.dateFormat = "HH:mm"
         return formatter.string(from: date)
     }
--- a/CriticalMass/FormatDisplay.swift
+++ b/CriticalMass/FormatDisplay.swift
@@ -46,6 +46,7 @@ enum FormatDisplay {
         let date = Date(timeIntervalSince1970: message.timestamp)
         let formatter = DateFormatter()
         formatter.calendar = calendar
+        formatter.timeZone = calendar.timeZone
         formatter.dateFormat = "HH:mm"
         return formatter.string(from: date)
     }
--- a/CriticalMass/FormatDisplay.swift
+++ b/CriticalMass/FormatDisplay.swift
@@ -31,9 +31,10 @@ private extension DateComponents {
 enum FormatDisplay {
     // inject current date to be able to keep the same time difference between tweet date and current date
     static var currentDate = Date()
+    static var calendar = Calendar.current
 
     static func dateString(for tweet: Tweet) -> String? {
-        let components = Calendar.current.dateComponents(
+        let components = calendar.dateComponents(
             [.second, .minute, .hour, .day, .month],
             from: tweet.created_at, to: currentDate
         ).dateComponentFromBiggestComponent
--- a/CriticalMass/Sequence+Additions.swift
+++ b/CriticalMass/Sequence+Additions.swift
@@ -4,9 +4,9 @@
 import Foundation
 
 extension Sequence {
-    func sorted<T: Comparable>(by keyPath: KeyPath<Element, T>, sortFunction: ((T, T) -> Bool) = (<)) -> [Element] {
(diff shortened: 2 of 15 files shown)


CriticalMass/FormatDisplay.swift
```diff
@@ -31,10 +31,11 @@ private extension DateComponents {
 enum FormatDisplay {
     // inject current date to be able to keep the same time difference between tweet date and current date
     static var currentDate = Date()
+    static var calendar = Calendar.current
```
lennet: Now that I look at it, I don't think its a good idea declare them statically because it would never change after launching the App right?

mltbnz: True. How about having the calendar as a default parameter in the signature?

CriticalMass/FormatDisplay.swift
```diff
@@ -31,10 +31,11 @@ private extension DateComponents {
...
 
     static func dateString(for tweet: Tweet) -> String? {
```
mltbnz:
```suggestion
    static func dateString(for tweet: Tweet, calendar: Calendar = .current) -> String? {
```

CriticalMass/FormatDisplay.swift
lennet: I moved both calendar and currentDate to the argument, but that breaks snapshots testing as they rely on that :/

CriticalMass/TwitterManager.swift
```diff
@@ -34,15 +34,14 @@ class TwitterManager {
 
     public func loadTweets(_ completion: ResultCallback<[Tweet]>? = nil) {
         networkLayer.get(request: request) { [weak self] result in
```
mltbnz: I think this early exit saves us from reading three lines with a nil coalescing operator and also assured that self is always there for the scope off the block

CriticalMass/TimeTraveler.swift
```diff
@@ -10,27 +10,35 @@ class TimeTraveler {
         self.date = date
     }
 
-    func travelTime(by interval: TravelInterval, _ calendar: Calendar = .current) {
+    @discardableResult
+    func travelTime(by interval: TravelInterval, _ calendar: Calendar = .current) -> Self {
```
mltbnz: This is odd. Looks like s builder pattern. But I can see that the API could be improved

CriticalMass/Sequence+Additions.swift
```diff
@@ -4,9 +4,9 @@
 import Foundation
 
 extension Sequence {
-    func sorted<T: Comparable>(by keyPath: KeyPath<Element, T>) -> [Element] {
-        sorted { a, b in
-            a[keyPath: keyPath] < b[keyPath: keyPath]
+    func sorted<T: Comparable>(by keyPath: KeyPath<Element, T>, sortFunction: ((T, T) -> Bool) = (<)) -> [Element] {
```
mltbnz: Maybe this should be called sortOperator and I think it could have a default implementation of < since that feels like the most natural ordering

CriticalMass/FormatDisplay.swift
```diff
@@ -45,9 +42,11 @@ enum FormatDisplay {
         return formatter.string(from: components)
     }
 
-    static func hoursAndMinutesDateString(from message: ChatMessage) -> String {
+    static func hoursAndMinutesDateString(from message: ChatMessage, calendar: Calendar = .current) -> String {
```
mltbnz: Could be moved to the DateFormatter extension

mltbnz:
In general I think the PR is not really focused and does s lot of changes in different classes. But to be fair I am struggling with that myself 😄

CriticalMass/TweetTableViewCell.swift
```diff
@@ -62,8 +62,13 @@ class TweetTableViewCell: UITableViewCell, MessageConfigurable, IBConstructable
...
     @IBOutlet private var nameAndHandleStackView: UIStackView!
 
+    // We need both setup methods to maintain conformance to the protocol + be able to inject a custom date for testing
     func setup(for tweet: Tweet) {
-        dateLabel.text = FormatDisplay.dateString(for: tweet)
+        setup(for: tweet, currentDate: Date())
+    }
+
+    func setup(for tweet: Tweet, currentDate: Date) {
+        dateLabel.text = FormatDisplay.dateString(for: tweet, currentDate: currentDate)
```
mltbnz: If we would have ViewModels I'd say we should actually move the String creation to that one but we could at least move it from the View to the Model and make it a computed property. Also you could use the TimeTraveler for the date generation

CriticalMass/TimeTraveler.swift
lennet: TimeTraveler was already kind of a Builder class before. This changes just adds the ability to chain the operations which makes it possible to use it in a single line to support property or argument default assignments. It's actually not necessary anymore because I removed the part that's relying on it in the meantime

CriticalMass/Sequence+Additions.swift
lennet: I agree with the default implementation, however we have currently more calls were we are using > and I don't want to further grow the scope of this PR. So I'll suggest keeping it as it is for this PR.

CriticalMass/FormatDisplay.swift
lennet: Yes, but that's maybe something for a future PR

CriticalMass/TweetTableViewCell.swift
lennet: Also something for a future PR

CriticalMass/TwitterManager.swift
lennet: > In general I think the PR is not really focused and does s lot of changes in different classes. But to be fair I am struggling with that myself 😄

I just wanted to write some tests and ended up with fixing bugs 🙈. To not further grow the scope of this PR, I address PR comments on Issues that are not caused by this PR. I hope that's okay

lennet:
> In general I think the PR is not really focused and does s lot of changes in different classes. But to be fair I am struggling with that myself 😄

I just wanted to write some tests and ended up with fixing bugs 🙈. To not further grow the scope of this PR, I address PR comments on Issues that are not caused by this PR. I hope that's okay

CriticalMass/TwitterManager.swift
lennet: Yes but if we have an active completion block we should always try to call it otherwise the calling function might wait forever and show an infinite loading animation for example.

CriticalMass/FormatDisplay.swift
mltbnz: Whoat? Sounds strange as to me this is the same

CriticalMass/TwitterManager.swift
mltbnz: Good point. I missed that one.
The Manager also handing out view states feels weird though

PR #267, opened 2020-03-10
Convert string extension for NSLocalizedString to L10n enum

mltbnz:
## 📲 What

Convert extension on `String` to `L10n` enum.

## 🤔 Why

I was unhappy with the `String.` call for a localized string since it somehow felt weird.
Also removes a little bit of boilerplate code.


CriticalMass/L10n.swift
```diff
@@ -0,0 +1,52 @@
...
+    static let chatSend = L10n.translate("chat.send")
+    static let chatTitle = L10n.translate("chat.title")
+    static let chatSendError = L10n.translate("chat.send.error")
+    // Map
+    static let mapLayerInfo = L10n.translate("map.layer.info")
+    static let mapLayerInfoTitle = L10n.translate("map.layer.info.title")
+    static let mapTitle = L10n.translate("map.title")
+    // ErrorState
+    static let errorStateTitle = L10n.translate("errorState.title")
+    static let errorStateMessage = L10n.translate("errorState.message")
```
fbernutz: i like that you can reduce the boilerplate `comment: ""` here

mltbnz:
I like omitting things but here I would favour a the L10n version since to me it better explains what it behind that value. With the string extension code completion is cluttered anyways and also its not clear to me what Kind of value hides behind

fbernutz:
> I like omitting things but here I would favour a the L10n version since to me it better explains what it behind that value. With the string extension code completion is cluttered anyways and also its not clear to me what Kind of value hides behind

True! 💯 That's an argument!

PR #271, opened 2020-03-17
Create stale.yml

mltbnz:
## 📲 What

Add `stale` action

## 🤔 Why

Since there are a lot of issues that are not being updated I think we should at least mark them to indicate that there is no recent process. There are some issues which are 3 years old and maybe should mark them or better close them since it does not seem anyone is interested in them anymore.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,19 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'Stale issue message'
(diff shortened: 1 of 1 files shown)


mltbnz:
I have no idea tbh.
Maybe adding some parameters would be a good idea. In general I just think these 5 year old issues won’t be tackled anymore and should be closed some time in the future. But first label them as not active. 

Oh and I think they will be levels but I’ll check on my firm first

fbernutz:
Proposal: Name the label simply "stale" and update the message to something more understandable, like "This issue/PR is stale because it has been open for X days with no activity. Remove the stale label or comment or this will be closed in X days." 

Inspired from [WeTransfer-iOS-CI](https://github.com/WeTransfer/WeTransfer-iOS-CI/blob/master/.github/workflows/stale.yml)

fbernutz:
We have some PRs labeled with `on hold`. We can't let the stale action ignore PRs with specific labels, right?

mltbnz:
Gonna close this. Think it is not really needed

mltbnz:
Thx for the Feedback nonetheless

PR #272, opened 2020-03-18
Create .gitattributes

mltbnz:
set gitattributes to help with merging project files

## 📲 What

Add `.gitattributes` with `*.pbxproj merge=union` attribute

## 🤔 Why

With this git is able to resolve some merge conflicts with the `.pbxproj` file itself. So big wins for everyone

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.pbxproj merge=union


mltbnz:
> Cool idea! Never worked with this before... Do you have experiences with this?

Yes. I can only say it sometimes works 🤣 If it does not you basically fall back to having a merge conflict 🤷‍♂️

fbernutz:
> > Cool idea! Never worked with this before... Do you have experiences with this?
> 
> Yes. I can only say it sometimes works 🤣 If it does not you basically fall back to having a merge conflict 🤷‍♂️

But it won't fix something wrong for me, right? If it's the case, let's try this out :)

mltbnz:
> > > Cool idea! Never worked with this before... Do you have experiences with this?
> > 
> > 
> > Yes. I can only say it sometimes works 🤣 If it does not you basically fall back to having a merge conflict 🤷‍♂️
> 
> But it won't fix something wrong for me, right? If it's the case, let's try this out :)

I think it does more good then bad
Further info:
https://roadfiresoftware.com/2015/09/automatically-resolving-git-merge-conflicts-in-xcodes-project-pbxproj/

PR #273, opened 2020-03-20
Add BikeAnnotation clustering

mltbnz:
## 📝 Docs

## 📲 What

Add `BikeClusterAnnotationView` and some minor fixes.

## 🤔 Why

Clustering a map is a good way of visually structuring and it makes the map easier to `read`.
The default zoomLevel is a good fit IMO. Also I consider this a low hanging fruit

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot 2020-03-20 at 15 19 52 | Screenshot 2020-03-20 at 14 56 06 |

### DIFF
--- a/CriticalMass/BikeAnnotationView.swift
+++ b/CriticalMass/BikeAnnotationView.swift
@@ -41,82 +41,10 @@ class BikeAnnoationView: MKAnnotationView {
         backgroundColor = .clear
 
         frame = Constants.shapeRect
-
         layer.addSublayer(ovalShapeLayer)
-    }
-}
 
-class FriendAnnotationView: MKAnnotationView {
-    var friendView: FriendView?
-
-    var friend: Friend? {
-        didSet {
-            friendView?.name = friend?.name ?? ""
+        if #available(iOS 11.0, *) {
+            clusteringIdentifier = BikeClusterAnnotationView.clusteringIdentifier
         }
-    }
-
-    override init(annotation: MKAnnotation?, reuseIdentifier: String?) {
-        super.init(annotation: annotation, reuseIdentifier: reuseIdentifier)
-        commonInit()
-    }
-
-    required init?(coder aDecoder: NSCoder) {
-        super.init(coder: aDecoder)
-        commonInit()
-    }
-
-    private func commonInit() {
-        canShowCallout = false
-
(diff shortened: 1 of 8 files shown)


CriticalMass/BikeAnnotationView.swift
```diff
@@ -41,82 +41,54 @@ class BikeAnnoationView: MKAnnotationView {
...
-        didSet {
-            friendView?.name = friend?.name ?? ""
+        if #available(iOS 11.0, *) {
+            clusteringIdentifier = BikeClusterAnnotationView.clusteringIdentifier
         }
     }
+}
+
+@available(iOS 11.0, *)
+class BikeClusterAnnotationView: MKAnnotationView {
```
fbernutz: the diff is super difficult to read 🙈 You basically added the `BikeClusterAnnotationView` and moved `FriendAnnotationView` and `FriendView` into an own file, right?

lennet:
How does it look like for a mass? Can you still see the snake?

mltbnz:
Sure. It’ll only cluster for far out zoom levels. I’ll add a gif to demonstrate

CriticalMass/BikeAnnotationView.swift
mltbnz: Yeah. That was probably not the best idea

mltbnz: Yeah. That was probably not the best idea

mltbnz: moved it the cluster annotation view to its own file 📁

mltbnz:
> How does it look like for a mass? Can you still see the snake if you zoom in?

But it could actually be a problem since the cluster is based on a collision rect. So you’d have some cluster here and there and others would be displayed as single dots. Maybe this wasn’t such a low hanging fruit. Just needed to move my thoughts away from this corona craziness.

mltbnz:
But I think this can be tweeked to only form a cluster when a certain zoom level was reached 🛠🔦

mltbnz:
So I think the mapviews clustering is not the right choice to give users an overview of how many people are attending. The collisionMode cant be tweaked or set via zoomLevel. I'll close this 🤷‍♂️

PR #276, opened 2020-03-25
Add context menu or actionsheet to next event annotation

mltbnz:
## 📲 What

Add `UIMenu` support for iOS 13 and a `UILongPressGestureRecognizer` for < iOS13 to add some context to the Critical Mass Annotation.

* also adds an AlertPresenter to consolidate alert presenting. Please excuse cause not really part of this task

## 🤔 Why

To better integrate in the OS and to add some functionality to the Event annotation

#### iOS 13
Screenshot 2020-03-25 at 11 33 08

#### iOS 10-12
Screenshot 2020-03-25 at 11 30 36

## ♿️ Accessibility 

- [x] Sends UIAccessability notification when actionSheet appears

### DIFF
--- a/CriticalMass/AlertPresenter.swift
+++ b/CriticalMass/AlertPresenter.swift
@@ -1,0 +1,32 @@
+//
+// Created for CriticalMaps in 2020
+
+import UIKit
+
+final class AlertPresenter {
+    private var viewController: UIViewController? {
+        UIApplication.shared.keyWindow?.rootViewController
+    }
+
+    static let shared = AlertPresenter()
+
+    private init() {}
+
+    func presentAlert(
+        title: String?,
+        message: String? = nil,
+        preferredStyle: UIAlertController.Style,
+        actionData: [UIAlertAction],
+        isCancelable: Bool = false
+    ) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: preferredStyle)
+        actionData.forEach(alert.addAction(_:))
+        if isCancelable {
+            let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+            alert.addAction(cancelAction)
+        }
+        onMain {
+            self.viewController?.present(alert, animated: true, completion: nil)
+        }
+    }
+}
--- a/CriticalMass/MapViewController.swift
+++ b/CriticalMass/MapViewController.swift
@@ -309,9 +309,7 @@ extension MapViewController: MKMapViewDelegate {
     }
 
     private func routeToEvent() {
-        nextRideManager.nextRide.flatMap {
-            $0.openInMaps()
-        }
+        nextRideManager.nextRide?.openInMaps()
     }
 
     private func shareEvent() {
--- a/CriticalMass/MapViewController.swift
+++ b/CriticalMass/MapViewController.swift
@@ -254,10 +254,14 @@ extension MapViewController: MKMapViewDelegate {
         // TODO: Remove workaround when target > iOS10 since it does not seem to work with the MKMapView+Register extension
         if annotation is CriticalMassAnnotation {
             if #available(iOS 11.0, *) {
-                return mapView.dequeueReusableAnnotationView(
+                let view = mapView.dequeueReusableAnnotationView(
                     withIdentifier: CMMarkerAnnotationView.reuseIdentifier,
                     for: annotation
                 )
+                let markerView = view as! CMMarkerAnnotationView
+                markerView.shareEventClosure = { self.shareEvent() }
+                markerView.routeEventClosure = { self.routeToEvent() }
+                return view
             } else {
                 return mapView.dequeueReusableAnnotationView(withIdentifier: CMAnnotationView.reuseIdentifier) as? CMAnnotationView
                     ?? CMAnnotationView(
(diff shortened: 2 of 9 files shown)


CriticalMass/AlertPresenter.swift
```diff
@@ -0,0 +1,33 @@
...
+    ) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: preferredStyle)
+        actionData.forEach(alert.addAction(_:))
+        if isCancelable {
+            let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
+            alert.addAction(cancelAction)
+        }
+        onMain {
+            self.viewController?.present(alert, animated: true, completion: nil)
+            UIAccessibility.post(notification: .announcement, argument: nil)
```
lennet: I don't think that is needed. UIAlertController does that automatically.

CriticalMass/MapViewController.swift
```diff
@@ -254,10 +254,15 @@ extension MapViewController: MKMapViewDelegate {
...
             if #available(iOS 11.0, *) {
-                return mapView.dequeueReusableAnnotationView(
+                let view = mapView.dequeueReusableAnnotationView(
                     withIdentifier: CMMarkerAnnotationView.reuseIdentifier,
                     for: annotation
                 )
+                if #available(iOS 13.0, *) {
+                    let interaction = UIContextMenuInteraction(delegate: self)
+                    view.addInteraction(interaction)
+                }
```
lennet: How about adding the Interaktion inside CMMarkerAnnotationView ?

CriticalMass/MapViewController.swift
```diff
@@ -280,6 +285,64 @@ extension MapViewController: MKMapViewDelegate {
...
+            preferredStyle: .actionSheet,
+            actionData: [routeAction, shareAction],
+            isCancelable: true
+        )
+    }
+
+    private func routeToEvent() {
+        nextRideManager.nextRide.flatMap {
+            $0.openInMaps()
+        }
```
lennet: Shouldn't we use forEach here since we don't want to create a new collection

CriticalMass/MapViewController.swift
```diff
@@ -280,6 +285,64 @@ extension MapViewController: MKMapViewDelegate {
...
+    }
+
+    private func shareEvent() {
+        nextRideManager.nextRide.flatMap {
+            let activityViewController = UIActivityViewController(
+                activityItems: [$0.shareMessage],
+                applicationActivities: nil
+            )
+            self.present(activityViewController, animated: true, completion: nil)
```
lennet: same here

CriticalMass/AlertPresenter.swift
mltbnz: Wasn’t sure if it does. Is there and place those things get documented?

CriticalMass/MapViewController.swift
mltbnz: I think i don’t get what you mean. But Maybe the GitHub app placed the comment wrong.

Where do you want to use it?

CriticalMass/MapViewController.swift
mltbnz: ?

CriticalMass/MapViewController.swift
mltbnz: Possible but to me this is controller code. But I don’t have a strong opinion about it 🤷‍♂️

CriticalMass/MapViewController.swift
lennet: Ah I though it is a collection, so my comment doesn't make sense..

Why are you using flatMap here and not simple optional changing?

CriticalMass/MapViewController.swift
lennet: Oh right. CMMarkerAnnotationController seems to be right class for that

CriticalMass/AlertPresenter.swift
lennet: I'm not aware of any documentation about the Accessibility Notifications. 

A good approach is to test the App first with assistive technologies and then add notifications if needed. Also the .announcement notification should only be used if you want that the screenreader announces text. If you want to notify the assistive tech that the visible UI elements changed you can use .layoutChanged or .screenChanged.

CriticalMass/MapViewController.swift
mltbnz: I meant ViewController code. moved it in the view now

CriticalMass/MapViewController.swift
mltbnz: good. call didn't really pay attention

CriticalMass/MapViewController.swift
lennet: It's also about the flatMap here. I'm not a fan of using it for unwrapping an optional (except for when it used to transform a collection into another collection)

mltbnz: But it’s totally valid code since it s not an operator for collections but rather for optionals and I think in a case where we are only interested if a value exists it is even improving readability.
What does you make not a big fan?

I can change it to`if let` but I’m not convinced yet by the argument’s

CriticalMass/AlertPresenter.swift
```diff
@@ -0,0 +1,32 @@
...
+        title: String?,
+        message: String? = nil,
+        preferredStyle: UIAlertController.Style,
+        actionData: [UIAlertAction],
+        isCancelable: Bool = false
+    ) {
+        let alert = UIAlertController(title: title, message: message, preferredStyle: preferredStyle)
+        actionData.forEach(alert.addAction(_:))
+        if isCancelable {
+            let cancelAction = UIAlertAction(title: "Cancel", style: .cancel)
```
fbernutz, line 25: The "Cancel" should be localized :)

PR #277, opened 2020-03-25
add MAKEFILE and update README

mltbnz:
## 📲 What

Add a Makefile to setup new devs easier.

## 🤔 Why

To add the tools needed for this repo the README was lacking a setup step.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+all: bootstrap
+
+bootstrap: dependencies
+
+dependencies:
+	$(shell ./update_dependencies.sh run_swiftformat)
+	$(shell ./update_dependencies.sh run_license_plist)
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@
 	<img src="https://badges.crowdin.net/critical-maps/localized.svg" alt="Localized Status" />
 </a>
 
+## Setup
+
+- Run `make bootstrap` to install tools and dependencies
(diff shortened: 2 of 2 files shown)


lennet:
Can you elaborate why it's needed? The tools are shipped with the repo and the current master should contain a formatted version. I would prefer a setup where no manual setup steps are needed and a contributor can just open the project in Xcode

mltbnz:
where is swiftformat for example shipped with the repo? Must have missed that.
If so why is it shipped with the repo?

lennet:
in the [bin folder](https://github.com/criticalmaps/criticalmaps-ios/tree/master/bin). The only tool we are not shipping with the repo is fastlane but that's also not required on the machine

mltbnz:
So it is there as a binary that is updated with the github action on a weekly basis?

lennet:
yes

mltbnz:
> yes

mltbnz:
But I guess it didn’t make much sense then.
I wasn’t aware there are binaries we keep in the repo or I forgot about it probably.

Is the Format build Phase actually using it?

I think it would be cool to have 3rd dependencies in the app via SPM and then use ` swift run` but I think for that the  critical maps kit would needed to be extracted first

mltbnz:
Postponed until a nicer solution is at hand

PR #282, opened 2020-04-07
Added syntatic sugar to some sequence methods

lennet:
## 📲 What

I started to replace {.$0.property } with the new key path functions introduced in swift 5.2 \.property and then experimented with some more helpers to replace $0 syntax with a more easy readable API and introduced `castTo` and `canBeCastedTo`.

## 🤔 Why

The goal of this PR is to make the code more readable. On the other side, we keep introducing custom helpers so it might also have the opposite effect. What do you think?


lennet:
Makes sense. I reverted the changes

Issue #190, opened 2019-11-04
Only reload cells that changed for MessagesTableViewController


lennet:
https://github.com/criticalmaps/criticalmaps-ios/blob/600f0853a6d30ea135ab5247fa04c281fec159e8/CriticalMass/MessagesTableViewController.swift#L29

We are currently just calling `reloadData` if we receive new message objects which is not super efficient. It would be nice if we would just update the rows that were inserted, deleted or updated.

iglushchuk:
Hi! I can handle it

lennet:
Cool!

iglushchuk:
Hi! I really apologise, that I didn't resolve that problem, recently I was really busy. Now I have time and I'd like to fix that issue in the near future :)
I have investigated how to solve this issue, I understand now how it can be fixed. Also I have a question: what do you think if we use new diffable datasource for iOS 13 and for iOS 10-12 we will keep reloadData?

lennet:
No worries! 

Sounds like a good Idea to me!

PR #294, opened 2020-06-21
Only reload cells that changed for MessagesTableViewController #190

iglushchuk:
This Pull Request fixes issue #190 

Here I added two different dataSources: one for iOS 13 using "Diffable data source" API and for previous iOS versions I made data source, which reloads with tableView.reloadData(). Messages storage moves to datasources, so now they are source of truth.

I was thinking about making a protocol, but that protocol should be generic, and I would have to use type erasure on this protocol, which looked like overengineering and I would loose type safety, provided by generic MessagesTableViewController. Also I didn't figure out how to unit test this data sources.

Looking forward to code review :)


CriticalMass/MessagesDefaultDataSource.swift
```diff
@@ -0,0 +1,37 @@
...
+    func tableView(_: UITableView, numberOfRowsInSection _: Int) -> Int {
+        messages.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(ofType: T.self)
+        cell.setup(for: messages[indexPath.row])
+
+        return cell
+    }
```
lennet, line 36: Maybe we can move that code to MessagesDataSource to share it with MessageDiffableDataSource

CriticalMass/ChatMessage.swift
```diff
@@ -7,7 +7,7 @@
 
 import Foundation
 
-public struct ChatMessage: Codable, Equatable {
+public struct ChatMessage: Codable, Equatable, Hashable {
```
lennet, line 10:
```suggestion
public struct ChatMessage: Codable, Hashable {
```

lennet:
> 
> [...]Also I didn't figure out how to unit test this data sources. [...]
> 

We could unit test by creating a pure tableview in the test case and attach it to the datasource and send some messages to the datasource and see if the row count and the content matches and updates. That would be nice but not super important 

Maybe a SnapshotTests with both DataSource would be nice just to see that everything works since we won't regularly test the default data source.

CriticalMass/MessagesTableViewController.swift
```diff
@@ -70,30 +53,16 @@ class MessagesTableViewController<T: IBConstructableMessageTableViewCell>: UITab
...
-        self.messages = messages
-    }
-
-    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(ofType: T.self)
-        cell.setup(for: messages[indexPath.row])
-        return cell
+        dataSource.messages = messages
+        updateNoMessageCountIfNeeded()
+        tableView.reloadData()
```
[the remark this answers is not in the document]
iglushchuk: You are right, I will fix that

CriticalMass/MessagesDefaultDataSource.swift
iglushchuk: Good idea, I'll do that

iglushchuk:
Fixed issues and added unit/snapshot tests, check it please 😃

iglushchuk:
I saw that animation on tweets screen was not that good, so I turned off animation in diffable if we populate tableview form zero to N cells. In other cases there should be animation.

CriticalMapsSnapshotTests/MessagesTableViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,40 @@
+//
```
lennet, line 1: Nice Idea to use Snapshot tests here 👏

PR #309, opened 2020-08-09
Update "Update Dependency" Action to create an issue instead

fbernutz:
## 📲 What

Update the "Update Dependency" Github Action to behave like a reminder bot and create an issue to remind us to update some dependencies manually. We use the following Github Action for this: https://github.com/peter-evans/create-issue-from-file

## 🤔 Why

The automated "Update Dependency" Github Action was not working on CI. Therefore we now use Dependabot to update bundler dependencies. We now have to think about updating `swiftformat`, `license-plist` and Xcode Swift Packages manually.


mltbnz:
Will this create new issues every time it is running?

fbernutz:
Currently it will create new issues, but an update should also be possible... We can use the `issue-number` to update a specific issue (see [here](https://github.com/peter-evans/create-issue-from-file)), but I don't know what happens if the issue with the specific issue number is closed or simply doesn't exist yet. 🤷‍♀️

PR #320, opened 2020-09-15
bump deployment target to iOS 12

mltbnz:
## 📝 Docs

## 📲 What

Bump minimum deployment target to iOS 12

## 🤔 Why

As agreed upon this bumps the projects deployment target to iOS 12

### DIFF
--- a/CriticalMass/NextRidesRequest.swift
+++ b/CriticalMass/NextRidesRequest.swift
@@ -7,8 +7,8 @@ struct NextRideQuery: Codable {
     let centerLatitude: CLLocationDegrees
     let centerLongitude: CLLocationDegrees
     let radius: Int
-    let year = Date.getCurrent(\.year)
-    let month = Date.getCurrent(\.month)
+    var year = Date.getCurrent(\.year)
+    var month = Date.getCurrent(\.month)
 }
 
 struct NextRidesRequest: APIRequestDefining {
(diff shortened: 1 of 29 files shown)


CriticalMass/NextRidesRequest.swift
```diff
@@ -7,8 +7,8 @@ struct NextRideQuery: Codable {
     let centerLatitude: CLLocationDegrees
     let centerLongitude: CLLocationDegrees
     let radius: Int
-    let year = Date.getCurrent(\.year)
-    let month = Date.getCurrent(\.month)
+    var year = Date.getCurrent(\.year)
```
fbernutz, line 10: Was this change intended?

fbernutz (approved):
Looks good for me. I'm reviewing from my phone so I depend on the green tests 👌😉 Is there maybe an information about the iOS version in the readme or so, which needs to be updated?

CriticalMass/NextRidesRequest.swift
mltbnz: yes. there was a warning which I wanted to rid of.
Also I think this is the way to go when declaring properties that have a default value on structs

mltbnz:
> Looks good for me. I'm reviewing from my phone so I depend on the green tests 👌😉 Is there maybe an information about the iOS version in the readme or so, which needs to be updated?

I haven't seen additional info in the README but maybe good if I check other docs again

PR #322, opened 2020-09-16
[Settings] add alternate app icons

mltbnz:
## 📝 Docs

## 📲 What

Adding a section to the settings to choose an alternate app icon

## 🤔 Why

For users to customise their app.

## 👀 See

Screenshot 2020-09-16 at 19 41 32
Screenshot 2020-09-16 at 19 41 38

## ♿️ Accessibility 

- ~[ ] Tap targets use minimum of 44x44 pts dimensions~
- [x] Works with VoiceOver
- [x] Supports Dynamic Type


mltbnz:
@zutrinken I made a simple list to choose. I you have something other in mind we can discuss here maybe

zutrinken:
Looks legit 👌

CriticalMass/AppIcon.swift
```diff
@@ -0,0 +1,35 @@
...
+
+enum AppIcon: String, CaseIterable {
+    case light, dark, rainbow, yellow, neon
+
+    var name: String {
+        String(rawValue.prefix(1).uppercased() + rawValue.dropFirst())
+    }
+
+    var image: UIImage? {
+        let documentsUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first!
```
fbernutz, line 15: Xcode tells me that this line is unused ;) 
```suggestion
```

fbernutz (approved):
Is there a specific reason why the alternate app icons are not included in the `Images.xcassets`? 
But all in all the code looks good for me and changing the app icon works in my tests in simulator as expected :) Really excited about this feature ;)

mltbnz:
> Is there a specific reason why the alternate app icons are not included in the `Images.xcassets`?
> But all in all the code looks good for me and changing the app icon works in my tests in simulator as expected :) Really excited about this feature ;)

So this is the first time I added such a feature and from what I have saw online I think it is not possible to use images from the catalog. But I can take a second look. Maybe there is a way

fbernutz:
> > Is there a specific reason why the alternate app icons are not included in the `Images.xcassets`?
> > But all in all the code looks good for me and changing the app icon works in my tests in simulator as expected :) Really excited about this feature ;)
> 
> So this is the first time I added such a feature and from what I have saw online I think it is not possible to use images from the catalog. But I can take a second look. Maybe there is a way

Ah okay, I've never implemented this myself as well, so I have no idea... 😄 When it's not working with asset catalog, it's fine for me like this as well.

PR #325, opened 2020-09-16
[Next ride event] add settings option to select event search radius

mltbnz:
## 📝 Docs

## 📲 What

A description of the change.

## 🤔 Why

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

## ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type


CriticalMaps.xcodeproj/project.pbxproj
```diff
@@ -1289,7 +1292,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "PLISTBUDDY=\"/usr/libexec/PlistBuddy\"\nINFO_PLIST=\"$CODESIGNING_FOLDER_PATH/Info.plist\"\nSETTINGS_PLIST=\"$CODESIGNING_FOLDER_PATH/Settings.bundle/Root.plist\"\nVERSION_NUMBER=\"`$PLISTBUDDY -c \\\"Print CFBundleShortVersionString\\\" \\\"$INFO_PLIST\\\"`\"\nBUILD_NUMBER=\"`$PLISTBUDDY -c \\\"Print CFBundleVersion\\\" \\\"$INFO_PLIST\\\"`\"\n\n$PLISTBUDDY -c \"Set :PreferenceSpecifiers:1:DefaultValue '$VERSION_NUMBER ($BUILD_NUMBER)'\" \"$SETTINGS_PLIST\"\n";
+			shellScript = "PLISTBUDDY=\"/usr/libexec/PlistBuddy\"\nINFO_PLIST=\"$CODESIGNING_FOLDER_PATH/Info.plist\"\nSETTINGS_PLIST=\"$CODESIGNING_FOLDER_PATH/Settings.bundle/Root.plist\"\nVERSION_NUMBER=\"`$PLISTBUDDY -c \\\"Print CFBundleShortVersionString\\\" \\\"$INFO_PLIST\\\"`\"\nBUILD_NUMBER=\"`$PLISTBUDDY -c \\\"Print CFBundleVersion\\\" \\\"$INFO_PLIST\\\"`\"\n\n$PLISTBUDDY -c \"Set :PreferenceSpecifiers:2:DefaultValue '$VERSION_NUMBER ($BUILD_NUMBER)'\" \"$SETTINGS_PLIST\"\n";
```
fbernutz, line 1295: what is this change from 1 to 2 for? 🤔

fbernutz: Ah nevermind, it's because the event search radius information is on top in the settings.

Settings.bundle/Root.plist
```diff
@@ -6,6 +6,30 @@
...
+			<string>PSMultiValueSpecifier</string>
+			<key>Title</key>
+			<string>Event Search Radius</string>
+			<key>Key</key>
+			<string>searchRadiusSetting</string>
+			<key>DefaultValue</key>
+			<integer>1</integer>
+			<key>Titles</key>
+			<array>
+				<string>10</string>
```
fbernutz, line 20: Just a number seems not 100% clear to me, maybe adding " km" after the number?

fbernutz (approved):
Looks good in my opinion. When testing this, there was one issue (I think). I was asked for location permission (it's irrelevant if I choose precise location or not) and then navigated in the Settings App to the CM Settings. The list of Search Radius had nothing selected. I thought the default is 20 but it was not displayed as checked. Seems like the `Default Value` is not working 🤔 

simulator_screenshot_A32FFEBE-4438-4E65-B153-5C1B5CEF7173

EDIT: I fixed the issue with the default value. 🙃

Settings.bundle/Root.plist
fbernutz: When adding " km" to the titles entries, the feature still works and shows the additional information in the UI. :+1: What do you think @mltbnz ?

mltbnz: think that makes sense. 
Since I saw that the API now also supports Event types (like Kidical Mass, Critical Mass and others) I thought to a selection screen for events you are interested in as well and then maybe it actually makes sense to move the option code somewhere inside the app. maybe a button on the map to configure the feature and maybe some people do not even want to see events on the map... 🤔 

WDYT?

fbernutz: Yes, sounds like this setting should move inside the app. 🤔

mltbnz:
I will not merge this one in favour of an in app solution since there is other stuff to configure and it makes more sense to me to have that directly in the app.

See #334

PR #331, opened 2020-09-30
[Refactor] facade userdefaults

mltbnz:
## 📲 What

This PR is intending to cut the Stores into smaller pieces and instead of injecting or directly referencing `UserDefaults` using Protocols to define what properties are accessible from an injected resource.

The `UserDefaults` then conform to these Protocols.

## 🤔 Why

* Clearer dependencies
* Removes UserDefaults from Tests

### DIFF
--- a/CriticalMass/ChatInputViewController.swift
+++ b/CriticalMass/ChatInputViewController.swift
@@ -23,9 +23,8 @@ final class ChatInputViewController: UIViewController, IBConstructable {
     }
 
     @IBOutlet private var sendButton: UIButton!
-
+    var themeController: ThemeController!
     weak var delegate: ChatInputDelegate?
-    private let themeController = ThemeController()
 
     override func viewDidLoad() {
         super.viewDidLoad()
--- a/CriticalMass/MapViewController.swift
+++ b/CriticalMass/MapViewController.swift
@@ -15,6 +15,28 @@ class MapViewController: UIViewController {
     private let nextRideManager: NextRideManager
 
     private let mapInfoViewController = MapInfoViewController.fromNib()
+    private lazy var annotationControllers: [AnnotationController] = {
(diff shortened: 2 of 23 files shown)


CriticalMass/ChatInputViewController.swift
```diff
@@ -23,9 +23,8 @@ final class ChatInputViewController: UIViewController, IBConstructable {
     }
 
     @IBOutlet private var sendButton: UIButton!
-
+    var themeController: ThemeController!
```
mltbnz, line 26: now gets injected from the AppController

CriticalMass/MapViewController.swift
```diff
@@ -15,33 +15,13 @@ class MapViewController: UIViewController {
     private let nextRideManager: NextRideManager
 
     private let mapInfoViewController = MapInfoViewController.fromNib()
-
```
mltbnz, line 18: this is just moving a couple lines of code

fbernutz: 👍

CriticalMassTests/RatingHelperTests.swift
```diff
@@ -19,71 +19,96 @@ class MockRatingRequest: RatingRequest {
...
+            ratingStorage: ratingsStorage,
             ratingRequest: MockRatingRequest.self,
             currentVersion: "MockVersion"
         )
         execute(times: 5, ratingHelper.onLaunch())
         XCTAssertEqual(MockRatingRequest.requestCounter, 1)
     }
 }
+
+struct RatingStorageMock: RatingStorage {
```
fbernutz, line 109: Maybe it's a good idea to make all of the mock structs in tests private so it's only available in the same file?

fbernutz (approved):
Looks good for me :)

CriticalMassTests/RatingHelperTests.swift
mltbnz: For some it is actually. Others are needed in more then one place

Issue #182, opened 2019-10-28
Appearance settings


zutrinken:
Add "use system preferences" as an appearance option. With this setting the app can benefit from the automatic switch function on sunset.

<table>
<tr>
<td>
<img src="https://user-images.githubusercontent.com/888679/67723328-f2501780-f9db-11e9-834e-e902575be08e.png" />
</td>
<td>
<img src="https://user-images.githubusercontent.com/888679/67723327-f2501780-f9db-11e9-92f4-99db4ad7c52e.png" />
</td>
</tr>
</table>

lennet:
It might be helpful to cherry pick SetttingsViewController and SettingsSection from the Friends feature branch as it adds support for navigating to view controllers from settings which is needed for the transition for the  Appearance Settings.

mltbnz:
What should the app present when the OS is < 13?

lennet:
Wasn't the idea to migrate to the official API at some point and remove dark mode support for pre iOS 13?

mltbnz:
Would make the code nicer but unsure if removing functionality from pre ios 13 users is a good idea

lennet:
Yes, that's true, but still an option we should keep in mind. Especially because our custom solution becomes harder to maintain over time and also more user upgrade to iOS 13

mltbnz:
I am totally up for it but since the current min deployment target is 10. .. removing something that has been a part of the app since some time is not needed. The solution that is in the app is definitely hard to maintain already especially when adding new UI. 

Right now I would only remove the current implementation if the min deployment target is iOS 13

lennet:
How about using something like that https://github.com/microsoft/FluentDarkModeKit, that also requires dropping iOS 10. Also looks super similar to the official API, so once we our target is iOS 13, migrating to the official API shouldn't be a lot of work

zutrinken:
Does supporting < 13 also include native dark mode maps or only the UI? If native dark mode map isn't included I'd drop the support.

lennet:
For pre iOS 13 we are using custom dark map tiles 
<img width="582" alt="Bildschirmfoto 2020-03-25 um 20 04 11" src="https://user-images.githubusercontent.com/7677738/77575529-f5d8c380-6ed3-11ea-901d-6e2208e6b9ee.png">

mltbnz:
It’s so dark

Freelenzer:
Hi, I added a first implementation of improved appearance settings.
[...]

PR #336, opened 2020-10-03
Added options for appearance settings (#182)

Freelenzer:
## 📝 Docs

## 📲 What

Added an implementation of https://github.com/criticalmaps/criticalmaps-ios/issues/182
It enables the user to set if he wants to use system appearance settings or dark/light mode only.
On iOS < 13 the option is a switch.

## 🤔 Why

Users can customise their appearance. Default is system so the app 
But it also gives the users the option to quickly change e.g. from dark to light mode so he can see the map better if he is in light environment.

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

## ♿️ Accessibility 

- [x] Tap targets use minimum of 44x44 pts dimensions
- [x] Works with VoiceOver
- [x] Supports Dynamic Type


fbernutz (commented):
Thanks for the contribution 🎉 
I just tested your implementation and it works perfectly for < iOS 13 and for >= iOS 13. Good job! 
Currently the snapshot tests fail in CI because we still use Xcode 11.5 and therefore iOS 13.7 in the CI. I added an issue for that, so we run tests in CI with iOS 14. https://github.com/criticalmaps/criticalmaps-ios/issues/337 
Until this is resolved, could you record the tests again with an iOS 13 simulator? Then it should be all green and we could merge this.

Freelenzer:
ok got it :)
Everything green now.

fbernutz (approved):

zutrinken:
Hey @Freelenzer thanks so much for your PR. I Finally had the time to check it out and found an issue that might relate to this. Maybe you wanna check it out: #379

fbernutz:
Just fixed with #380

PR #338, opened 2020-10-04
Maintenance

fbernutz:
## 📲 What

- Update `.gitignore` to keep track of `Package.resolved` file.
- Use non-deprecated Bundler Configuration and add configuration file for project
- Update Swift Packages to get support for Xcode 12 for Snapshot Testing
- Update Dependencies
- Run Swiftformat

This won't update the CI and tests to use Xcode 12. There's an own issue for that.

## 🤔 Why

It's good to see on which version a SPM is currently used and maintenance is always good ;)


.bundle/config
```diff
@@ -0,0 +1,2 @@
+---
+BUNDLE_PATH: vendor/bundle
```
fbernutz, line 2:
```suggestion
BUNDLE_PATH: vendor/bundle

```

mltbnz (approved):

PR #339, opened 2020-10-06
FIX Dark mode map overlay on > iOS 13

mltbnz:
+ do not overlay map with custom tiles when > ios13

fixes custom tile overlay introduced in #336


fbernutz (approved):
makes sense 👍

Issue #329, opened 2020-09-25
Legal button is behind other buttons


jacquealvesb:
The legal button is behind the other buttons, so we can't see it. Maybe putting it on the info section on settings would be a good option.

![IMG_1712](https://user-images.githubusercontent.com/5198967/94320951-a8154480-ff64-11ea-93c5-ae3dcf5ff0e5.PNG)

iPhone 7, iOS 14.0.1

mltbnz:
fixed in #342

PR #342, opened 2020-10-08
fix MavView `bottomContentOffset` (for legal info)

thisIsTheFoxe:
## 📝 Docs

## 📲 What

Fix the bottom (y-) positioning of the NavigationOverlay 'tab/nav bar'.

Alternative (suggested by issue-creator):
Which I don't think will work with the AppStore guidelines...
> Maybe putting it on the info section on settings would be a good option.

## 🤔 Why

The **center** of the NavOverlay was set to the height of the 'tab/nav-Bar', causing a margin half the overlays height. Therefore the highest point of the overly is exactly 1.5 * the overlays height from the bottom of the screen. This was ignored when setting the `bottomContentOffset`, blocking the "Legal" button and potentially causing further problems in future updates.
Furthermore, `bottomContentOffset` takes into account `layoutMargins.bottom`. Therefore the "Legal" button was only blocked on iPhone != X. Moving the NavOverlay a bit further up on those devices, improves usability.
> **Avoid explicitly placing interactive controls at the very bottom of the screen and in corners.** People use swipe gestures at the bottom edge of the display to access features like the Home screen and app switcher, and these gestures may cancel custom gestures you implement in this area. The far corners of the screen can be difficult areas for people to reach comfortably.
-Apple

## 👀 See

#329

| Before 🐛 | After 🦋 |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2020-10-08 at 20 46 19 | Simulator Screen Shot - iPhone 8 - 2020-10-08 at 20 44 52 |
## ♿️ Accessibility 

- [x] Works with VoiceOver
- [ ] Supports Dynamic Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Changelog for Critical Maps iOS
 - Fix Dark Mode Color Behaviour in Action Indicator (in Rules, Settings) and Separator (in Settings)
 - Fix missing information for VoiceOver user for the navigationbar buttons
 - Fix VoiceControl labels on each view
+- Fix `bottomContentOffset` for NavigationOverlay (did hide the Legal button)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


mltbnz (approved):
Thanks for the PR.
Looks good I think.

After this was mentioned I was a little surprised that this never was mentioned during app review 🤷‍♂️ But we are good now

PR #345, opened 2020-10-11
Update localisations

mltbnz:
## 📲 What

Update the `.strings` files in the project with the latest from the https://crowdin.com/project/critical-maps
Also this adds a `Generated` folder with generated Swift code for Localisation strings and images. 

## 🤔 Why SwiftGen?

+ acces type safe images
+ use generated code to make life a little easier and more expressive

## Questions

+ Do we want to add SwiftGen as a Dependency? Though it would then maybe need a custom update step too

### Also ➕

* minor things I think that make sense for snapshot testing. 
📌 is to check when executing if the device is running iOS 14 to avoid recording with another version then the simulator


mltbnz:
Really no idea why the snapshot tests keep failing...
They also failed in Xcode with some very minor differences
Screenshot 2020-10-13 at 16 46 25
Thats why I updated all the snapshots... Is there an Xcode 13 version which is causing issues?

and in another project it was the `precision` value which here is also very high. I felt the sweet spot was `0.9`
Maybe @fbernutz  you have an idea?

mltbnz:
> Really no idea why the snapshot tests keep failing...
> 
> They also failed in Xcode with some very minor differences
> 
> 
> Thats why I updated all the snapshots... Is there an Xcode 13 version which is causing issues?
> 
> 
> 
> and in another project it was the `precision` value which here is also very high. I felt the sweet spot was `0.9`
> 
> Maybe @fbernutz  you have an idea?
> 
> 

Found it

swiftgen.yml
```diff
@@ -0,0 +1,13 @@
...
+ outputs:
+   - templateName: structured-swift4
+     output: CriticalMass/Generated/Strings.swift
+
+xcassets:
+  - inputs: 
+    - CriticalMass/Images.xcassets
+    outputs:
+      - templateName: swift4
+        output: CriticalMass/Generated/Images.swift
```
fbernutz, line 13:
```suggestion
        output: CriticalMass/Generated/Images.swift
        
```

always end your file with a newline 🙃

fbernutz (commented):
Looks good! I've never worked with SwiftGen before, so I'm really curious how this will help us and how it feels like to use it ;) So, just to be clear, the master for the strings is now the `Localizable.strings` and from there the `Strings.swift` is generated, right? Can the generation of the resources be a Build Phase, like Swiftformat? Or can we add a CI check to see if we didn't forget to run SwiftGen? 🤔

fbernutz:
You mean #347 right? ;) Let's try this out! Seems like a good idea! I think automation can wait until we see a benefit in it.

mltbnz:
> You mean #347 right? ;) Let's try this out! Seems like a good idea! I think automation can wait until we see a benefit in it.

Right 😄
I think so too that Automation can come in later

fbernutz (approved):

PR #346, opened 2020-10-11
Update tweet cell datelabel timeStyle to abbreviated

mltbnz:
## 📲 What

+ Update the `timeStyle` in the TwitterCell DateFormatter from `.short` to **`.abbreviated`**
+ Uppercase the abbreviated timeLabel

## 🤔 Why

In some languages the `.short` timeStyle of the `DateComponentsFormatter` produces some long strings e.g. in Italian 1 Day becomes **1 giorno** In the case of the tweet date time interval it is more appropriate to display an `abbreviated` timeStyle since we save space and it is common to do. Also Twitter itself uses that format 

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
| Screenshot 2020-10-11 at 20 05 37 | Screenshot 2020-10-11 at 20 06 50 |

### Also

* SnapShot and Unit test xctesplans are updated with a pinned `region` and `language` for reproducible tests


fbernutz (approved):
👍

PR #347, opened 2020-10-13
[Snapshot tests] add locale configurations to tesplan to snapshot test localisations

mltbnz:
## 📝 Docs

## 📲 What

+ Add locale configurations to Snapshot testplan
+ Add `Locale` to snapshot test name

## 🤔 Why

This will execute Snapshot tests for all configured locales the app is using. That way we can make sure that updates do not lead to misconfigured views since missing sync keys in localisation files will pick the desired key as text label.


swiftgen.yml
```diff
@@ -10,5 +10,4 @@ xcassets:
     - CriticalMass/Images.xcassets
     outputs:
       - templateName: swift4
-        output: CriticalMass/Generated/Images.swift
-        
+        output: CriticalMass/Generated/Images.swift
```
fbernutz, line 13: was this change intended?

CriticalMapsSnapshotTests/CriticalMapsSnapshotTests.xctestplan
```diff
@@ -2,13 +2,48 @@
   "configurations" : [
     {
       "id" : "9675AA6D-95DA-4EBD-985A-7278586C91C6",
-      "name" : "Default",
+      "name" : "de_DE",
+      "options" : {
+        "language" : "de",
+        "testExecutionOrdering" : "random"
```
fbernutz, line 8: Some configurations have a random test execution order and some not. Do we want to add it to all configurations?

fbernutz (commented):
I think this is a nice idea to catch missing translations! 
There are some unused snapshot references without the locale in the file name which can be deleted. 
The italian EventSettings Snapshot `testGeneralAppearanceWhenNotificationsAreEnabled-it_IT-current.Dark` and the light version as well display some weird search radius information... Do they need a "simulated location"? Why is it only in the italian version?

CriticalMapsSnapshotTests/CriticalMapsSnapshotTests.xctestplan
mltbnz: oh. think that is a good idea

swiftgen.yml
mltbnz: I do not get that change tbh. I think I reverted your commit by it so could you explain that one again? 4726ef0262ec0d7581f26b2adc90d0477e4c6a15

fbernutz: Sure! I've added the newline because according to POSIX, every text file should end with a \n, or “newline” character.

You can read more about it here: https://stackoverflow.com/a/729795 

Maybe it's just nitpick...

In GitHub (only in the browser) the missing newline is marked with a red icon.

mltbnz: Interesting. I'll revert that

mltbnz:
> I think this is a nice idea to catch missing translations! 
> 
> There are some unused snapshot references without the locale in the file name which can be deleted. 
> 
> The italian EventSettings Snapshot `testGeneralAppearanceWhenNotificationsAreEnabled-it_IT-current.Dark` and the light version as well display some weird search radius information... Do they need a "simulated location"? Why is it only in the italian version? 

Ok haven't noticed that. Will take a look

fbernutz (approved):
looks good now :)

PR #351, opened 2020-10-17
[Map] add server update errorview on map

### LINKED ISSUE #268: Display Error message if we can't update the Map
Bildschirmfoto 2020-03-10 um 14 01 09

We want to display an Error message if we are unable to update the data. Parts of the feature like UI are already implemented but the connection between UI and the RequestManager is missing.

Might be useful to implement #245 first

mltbnz:
## 📝 Docs

## 📲 What

Add Error indication view on `MapInfoViewController` when the server does respond with something other then `OK`

## 🤔 Why

To display the users when there the map currently does not get updated because the server might be down or some other error occurred.

## 👀 See

Screenshot 2020-10-17 at 20 37 19
Screenshot 2020-10-17 at 20 37 11

## ✅ Testing

Not a lot of testing happening here since I think once again the architecture gets in the way. Maybe a TODO from this is to find a new architecture for the app. MVP could be a solution. It is easy to implement and scales well. But maybe it is fine like it is... 

## ♿️ Accessibility 

- [x] Supports VoiceOver
- [x] Supports Dynamic Type 

## ⚠️ Issue 
closes #268


mltbnz:
@zutrinken I used a different design than the one from the linked issue. I thought that having a banner kind of thing all the time on top of the view would be a little too distracting and this one is a little lighter. WDYT?

CriticalMass/Generated/Images.swift
```diff
@@ -2,11 +2,11 @@
 // Generated using SwiftGen — https://github.com/SwiftGen/SwiftGen
 
 #if os(macOS)
-    import AppKit
```
fbernutz, line 5: What happened to the formatting in the generated files? Should they use 4 white spaces or 2 as indentation?

mltbnz: The `Generated` folder was not excluded from `SwiftFormat` which runs in every build...
I'll fix it

mltbnz: But in general I think that those are files we can ignore when it comes to coding conventions since they are autoGenerated

mltbnz: the checked in file on `master` now is formatted and this one is not and will also add an `exclude` flag to the swiftformat yaml. Another option is to autogenerate the files and then execute swiftFormat after that. From a script or something like https://github.com/yonaskolb/Beak

fbernutz: Ah, yeah... I don't think that the generated files should play any role in code conventions, but it's a good idea to exclude it for swiftformat so it's not "reformatted" every time. Was just curious about the changes here..

CriticalMass/MapInfoViewController.swift
```diff
@@ -117,6 +140,20 @@ class MapInfoViewController: UIViewController, IBConstructable {
...
+        guard
+            let result = notification.object as? ApiResponseResult,
+            result.isError().0 != showServerError
+        else {
+            return
+        }
+        let errorMessage = (result.isError().1 as! NetworkError).errorDescription ?? L10n.Map.Layer.Info.errorMessage
+        errorView.setErrorLabelMessage(errorMessage)
+        showServerError = result.isError().0
+        UIAccessibility.post(notification: .layoutChanged, argument: nil)
```
fbernutz, line 154: I think this should rather be an announcement with the error message as body. But I'm not 100% sure. 

```suggestion
        UIAccessibility.post(notification: .announcement, argument: errorMessage)
```

fbernutz: we can leave it like that, because all other Accessibility posts in this view controller use `.layoutChanged` as well.

CriticalMass/Result+Additions.swift
```diff
@@ -0,0 +1,24 @@
...
+    func get() throws -> Success {
+        switch self {
+        case let .success(value):
+            return value
+        case let .failure(error):
+            throw error
+        }
+    }
+
+    func isError() -> (Bool, Error?) {
```
fbernutz, line 16: I'm not sure about this tuple. The call side is a bit unexpected with `result.isError().0` and `result.isError().1`. Maybe we could split them in `var error: Error?` and `hasFailed: Bool`. What do you think about that?

CriticalMass/MapErrorView.xib
```diff
@@ -0,0 +1,76 @@
...
+                                        <rect key="frame" x="0.0" y="0.0" width="18" height="30"/>
+                                        <color key="tintColor" systemColor="systemYellowColor"/>
+                                        <constraints>
+                                            <constraint firstAttribute="height" constant="30" id="4rT-59-env"/>
+                                            <constraint firstAttribute="width" constant="18" id="Mvr-o7-SZJ"/>
+                                        </constraints>
+                                    </imageView>
+                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="ServerError" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontForContentSizeCategory="YES" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="xGR-b9-PgM">
```
fbernutz, line 35: Maybe we should add a `numberOfLines = 0` for this label. We set the error message in this label dynamically and with dynamic type we could reach a display limit with one line.

CriticalMass/MapInfoViewController.swift
```diff
@@ -117,6 +140,20 @@ class MapInfoViewController: UIViewController, IBConstructable {
...
     }
+
+    @objc private func didReceiveLocationsUpdate(notification: Notification) {
+        guard
+            let result = notification.object as? ApiResponseResult,
+            result.isError().0 != showServerError
+        else {
+            return
+        }
+        let errorMessage = (result.isError().1 as! NetworkError).errorDescription ?? L10n.Map.Layer.Info.errorMessage
```
fbernutz, line 151: I received a `fatalError` when I tried to play around with the error view. I disconnected the device from internet connection and connected it again. It then crashed because `result.isError().1` is nil in that case. There is no error any more. So this should be better: 
```suggestion
        let errorMessage = (result.isError().1 as? NetworkError)?.errorDescription ?? L10n.Map.Layer.Info.errorMessage
```

fbernutz (commented):
Sorry it took me so long to review it! I think this looks good! I like that the error view is so subtle. Added some minor and some more important comments 😉 

I think the contrast for the light mode error view still needs some optimization.. maybe @zutrinken can help us out here with an idea on how to improve the contrast? It looks good in dark mode! :) 

| light compact | light expanded | dark compact | dark expanded |
|----|----|----|----|
| Simulator Screen Shot - iPhone 11 Pro - 2020-10-29 at 08 28 24 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-29 at 08 28 27 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-29 at 08 24 33 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-29 at 08 24 39 |

CriticalMass/Result+Additions.swift
mltbnz: think that improves it. I was just a litte lazy I guess. I'll update the signature then to have named parameters or do you think this deserves its own Type?

mltbnz:
> Sorry it took me so long to review it! I think this looks good! I like that the error view is so subtle. Added some minor and some more important comments 😉
> 
> I think the contrast for the light mode error view still needs some optimization.. maybe @zutrinken can help us out here with an idea on how to improve the contrast? It looks good in dark mode! :)
> 

no worries.
That light version does look a little bit `light` Maybe just by using a bold font we could make it visible a bit better

CriticalMass/Result+Additions.swift
fbernutz: I don't think named parameters will solve the call side problematic `result.isError().isError` and `result.isError().error`... I would split the function into two functions or two separate variables to make it more clear what it's about.

fbernutz:
> That light version does look a little bit `light` Maybe just by using a bold font we could make it visible a bit better

I'm mostly worried about the yellow icon on the bright map. Maybe we can make the "!" black and not transparent. 🤔

zutrinken (commented):
Beside the error message itself it's recommended to give an action. In this case it's "Please check/come back later". Else we leave the user with a non working map and no idea what to do about it. So the size of the box has to be bigger to fit a proper instruction.

Also the subtle style may work at a desk. Many people check the app while doing a super quick stop or even while riding. A strong "in your face" style makes it easy to quickly realise that something isn't working. Even without reading. So I recommend the bolder red style as in the design.

CriticalMass/MapErrorView.xib
mltbnz: > Beside the error message itself it's recommended to give an action. In this case it's "Please check/come back later". Else we leave the user with a non working map and no idea what to do about it. So the size of the box has to be bigger to fit a proper instruction.
> 
> Also the subtle style may work at a desk. Many people check the app while doing a super quick stop or even while riding. A strong "in your face" style makes it easy to quickly realise that something isn't working. Even without reading. So I recommend the bolder red style as in the design.

Since it stays there until there is no error any more I thought it is nicer to have a subtle warning.
I can also make another version with the bigger message and we could A/B test it

mltbnz:
> Beside the error message itself it's recommended to give an action. In this case it's "Please check/come back later". Else we leave the user with a non working map and no idea what to do about it. So the size of the box has to be bigger to fit a proper instruction.
> 
> Also the subtle style may work at a desk. Many people check the app while doing a super quick stop or even while riding. A strong "in your face" style makes it easy to quickly realise that something isn't working. Even without reading. So I recommend the bolder red style as in the design.

Since it stays there until there is no error any more I thought it is nicer to have a subtle warning.
I can also make another version with the bigger message and we could A/B test it

mltbnz:
> > That light version does look a little bit `light` Maybe just by using a bold font we could make it visible a bit better
> 
> I'm mostly worried about the yellow icon on the bright map. Maybe we can make the "!" black and not transparent. 🤔

Just tested that and I like it

CriticalMass/MapErrorView.xib
mltbnz: currently I have to find a solution for the container that is not growing... But yes. would make sense I guess

CriticalMass/Result+Additions.swift
mltbnz: I splitted the two up

zutrinken:
"Wait and check back later" might be obvious for a tech savvy person. But not for everyone. That also raises the question "do you have to close or kill the app to make it work"? A clear guidance removes uncertainty and frustration.

I can't imagine any scenario to properly A/B test this. All we know is that the app is heavily used in action. I like to compare it to car interfaces, where you get bold, bright and colourful warning messages including sound.

mltbnz:
I have implemented the red banner version now. It also works so maybe then lets go with this one
Screenshot 2020-11-02 at 18 27 34

CriticalMapsSnapshotTests/XCTestCase+SnapshotTests.swift
```diff
@@ -31,7 +32,7 @@ extension XCTestCase {
                            as: .image(precision: precision, size: size),
                            named: theme.displayName,
```
fbernutz, line 33: because the theme display name is now localized, we get snapshot references with `*.Hell.png` and `*.Dunkel.png`. I'm thinking about using the raw value here instead of the display name. What do you think?
Otherwise we have to update snapshot references each time the localization for the theme display name changes.
```suggestion
                           named: theme.rawValue,
```

mltbnz: Oh. Well you don't want that. The rawvalue makes sense. The other one should only be used for screen appearance

mltbnz: Merged this from the app. Maybe snapshots need an update too

fbernutz (approved):

PR #357, opened 2020-11-02
display background location indicator

mltbnz:
## 📲 What

Set `showsBackgroundLocationIndicator` to true

## 🤔 Why

To be more transparent to users
https://developer.apple.com/documentation/corelocation/cllocationmanager/2923541-showsbackgroundlocationindicator


fbernutz (approved):
really cool! Why is this not the standard? 🤔

PR #359, opened 2020-11-03
Update RequestManager TimeInterval

mltbnz:
## 📝 Docs

## 📲 What

A description of the change.

## 🤔 Why

## 👀 See

| Before 🐛 | After 🦋 |
| --- | --- |
|  |  |

## ♿️ Accessibility 

- [ ] Works with VoiceOver
- [ ] Supports Dynamic Type 

#143


mltbnz:
This should address #143 but I am not sure if it does since the issue is not fully clear to me... From the Background the app does not send any location update afaik, right? At least from using Proxyman it does not look like it

mltbnz:
lets see when CMs happen again if this can be improved

PR #380, opened 2021-01-07
Fix Switching Appearance Settings and using `.system`

### LINKED ISSUE #379: Map doesn't update after appearance change
Changing appearance in app from custom back to system doesn't update the map.

***

### OS settings: `Light`

Screenshot 2021-01-06 at 22 03 09

Go to appearance:

1. switch to `Dark`
2. switch to `System`

Close settings. Map is still `Dark`

***

### OS settings: `Dark`

Screenshot 2021-01-06 at 22 04 22

Go to appearance:

1. switch to `Light`
2. switch to `System`

Close settings. Map is still `Light`

fbernutz:
## 📝 Docs

## 📲 What

Setting `overrideUserInterfaceStyle = .unspecified` "causes the view controller to inherit the interface style from the system or a parent view controller" which is exactly what we want in this case.

Tested the feature on iOS 14.2 and iOS 12.1 in simulator.

## 🤔 Why

Fixes #379 and simplifies the logic a lot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changelog for Critical Maps iOS
 ### Fixed
 
 - Adapt BikeAnnotation Size to equal the Sizes in Maps
+- Fix Switching Appearance Settings and using `.system`
 
 # [3.9.1] - 2020-11-20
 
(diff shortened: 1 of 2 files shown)


zutrinken (approved):
Works as expected 👍

mltbnz (approved):
Cool. One line to rule then

PR #389, opened 2021-03-01
Update Schedule of Dependabot

fbernutz:
## 📲 What

This changes the schedule to monthly for all dependabot updates. This also removes some unhelpful comments.

## 🤔 Why

I think monthly updates are fine enough.


mltbnz (approved):
📅

PR #390, opened 2021-03-08
[Events feature] add ride `enabled` parameter

mltbnz:
## 📝 Docs

## 📲 What

* Add `enabled` parameter to ride object. https://github.com/criticalmass-one/criticalmass-in/blob/19251cdcea9db966a5e411c8a058d7a04aa92c7f/src/Entity/Ride.php#L315
* Make RideManager check if the "nextRide" is enabled and otherwise do not save it as the next ride
* save nextRide it is enabled but does not have an eventType set. Might happen since the API is not 💯

## 🤔 Why

* we do not want to show cancelled events.
* rides without eventType were showing on the map as an annotation and also the event banner is shown but tapping on it did not focus the next event since it was not set in the 'RideManager'. Coming back to the codebase after 3 months (even 3 months of not coding this feels weird... but 🤷‍♂️)

### DIFF
--- a/CriticalMassTests/NextRideManagerTests.swift
+++ b/CriticalMassTests/NextRideManagerTests.swift
@@ -198,7 +198,8 @@ class NextRideManagerTests: XCTestCase {
             switch result {
             case .success:
                 XCTFail()
-            case .failure:
+            case .failure(let error):
+                XCTAssertEqual(error as! EventError, EventError.rideDisabled)
                 break
             }
             exp.fulfill()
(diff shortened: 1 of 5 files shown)


CriticalMassTests/NextRideManagerTests.swift
```diff
@@ -173,6 +173,40 @@ class NextRideManagerTests: XCTestCase {
...
+                )
+            )
+        )
+        // when
+        let exp = expectation(description: "Wait for response")
+        nextRideManager.getNextRide(around: CLLocationCoordinate2D.TestData.alexanderPlatz) { result in
+            switch result {
+            case .success:
+                XCTFail()
+            case .failure:
```
fbernutz, line 201: We could check for the specific error `EventError.rideDisabled` here, right? I think it would add more value to the test and would be more consistent to the other existing tests. 

```suggestion
            case .failure(let error):
                XCTAssertEqual(error as! EventError, EventError.rideDisabled)
```

fbernutz (approved):
Looks good to me!

PR #392, opened 2021-04-17
🚑 Ride type fixes that hopefully end the crashes

mltbnz:
## 📝 Docs

## 📲 What

Set `latitude` -> [BE property](https://github.com/criticalmass-one/criticalmass-in/blob/512f411a62a8053193464b5d0a1f84abbc98eb46/src/Entity/Ride.php#L156) 
and `longitude` -> [BE property](https://github.com/criticalmass-one/criticalmass-in/blob/512f411a62a8053193464b5d0a1f84abbc98eb46/src/Entity/Ride.php#L166)
as optional properties

By now most properties are optional and I even thought about making all of them optionals. If anyone is reviewing please check with the [Ride.php](https://github.com/criticalmass-one/criticalmass-in/blob/master/src/Entity/Ride.php) if maybe another of the properties in `Ride.swift` has to be optional... But I think I got them now. 👀

## 🤔 Why

Although I never any crash reports I think this should solve the issue of a crashing app 🤞

### DIFF
--- a/CriticalMassTests/NextRidesRequestTests.swift
+++ b/CriticalMassTests/NextRidesRequestTests.swift
@@ -64,7 +64,7 @@ class NextRidesRequestTests: XCTestCase {
         XCTAssertEqual(expectedRides, rides)
     }
 
-    func testParseRespone2() throws {
+    func testParseRespone_withLatAndLng() throws {
         let responseData = """
         [{
             "id":8091,
--- a/CriticalMassTests/NextRidesRequestTests.swift
+++ b/CriticalMassTests/NextRidesRequestTests.swift
@@ -6,11 +6,15 @@ import MapKit
 import XCTest
 
 class NextRidesRequestTests: XCTestCase {
+    let request = NextRidesRequest(
+        coordinate: CLLocationCoordinate2D(latitude: 42, longitude: 42),
+        radius: 21
+    )
+
     func testMakeRequest() throws {
         let month = Date.getCurrent(\.month)
         let year = Date.getCurrent(\.year)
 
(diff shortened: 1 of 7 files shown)


CriticalMassTests/NextRidesRequestTests.swift
```diff
@@ -6,24 +6,263 @@ import MapKit
...
+            disabledReasonMessage: nil,
+            rideType: nil
+        )
+
+        let expectedRides = [ride]
+        let rides = try request.parseResponse(data: responseData)
+        XCTAssertEqual(expectedRides, rides)
+    }
+
+    func testParseRespone2() throws {
```
fbernutz, line 67: Suggestion to update the test name
```suggestion
    func testParseRespone_withLatAndLng() throws {
```

CriticalMassTests/NextRidesRequestTests.swift
```diff
@@ -6,24 +6,263 @@ import MapKit
...
+                disabledReasonMessage: nil,
+                rideType: .criticalMass
+            )
+        ]
+        let rides = try request.parseResponse(data: responseData)
+        XCTAssertEqual(expectedRides, rides)
+    }
 
-        let request = NextRidesRequest(coordinate: CLLocationCoordinate2D(latitude: 42, longitude: 42), radius: 21)
+    func testParseRespone_withNonUniformObjectsinResponse() throws {
```
fbernutz, line 193:
```suggestion
    func testParseRespone_withNonUniformObjectsInResponse() throws {
```

fbernutz (approved):
Looks good to me! Doesn't look like you missed any additional optional properties. Hope this fixes the crash 🤞

mltbnz:
If it didn't fix it I'll disable the feature

PR #393, opened 2021-04-20
Fix flaky tests

mltbnz:
fixes flaky tests


mltbnz:
also makes them faster

PR #403, opened 2021-06-12
setup for rebuild

mltbnz:
## 📲 What

This is a rebuild of the app 😳
The app code works but it is getting harder and harder to find and fix things since it became quite a patchwork over the years. This is totally fine and just natural I think but to keep the app running in the long run I think it is a good idea to rebuild it with a proper data flow and modularisation in mind.
 
## 🤔 Why

+ data flow
+ dark mode
+ modules
+ TCA
+ SwiftUI


mltbnz:
One idea to push the develop branch when an example feature is implemented (working on the nextride feature for now) and from there onwards do PRs to develop

PR #442, opened 2022-05-15
Add all upcoming ride events list

mltbnz:
## 📝 Docs

## 📲 What

A user can now view all upcoming events from a list. The events will also be visible on the map. 

## 🤔 Why

Some users requested the feature and I also thought about it but didn't found the time.

## 👀 See

Screenshot 2022-05-15 at 10 04 45
Screenshot 2022-05-16 at 08 08 04

## ♿️ Accessibility 

- [x] Works with VoiceOver
- [x] Supports Dynamic Type


mltbnz:
@zutrinken Any thoughts on the design?

PR #469, opened 2023-04-24
Fix typos

woxtu:
## 📲 What

This PR fixes typos in the code.


mltbnz:
Thanks for another one.
I'll run them later on my local machine just to be sure but they all look good. 

Seems ci isn't working

PR #475, opened 2023-06-01
Adopt new api architecture

mltbnz:
## 📝 Docs

## 📲 What

Update api service to use new endpoints and architecture

## 🤔 Why

The api needed to be updated and is now using a new architecture.

**note**
Tests will break after merging this and since I was not happy with the app core test suite I decided there will be an updated one


CriticalMapsKit/Sources/ApiClient/Endpoint.swift
```diff
@@ -3,20 +3,34 @@ import Foundation
...
+      return baseUrl
+    }
+    let path = pathComponents.joined(separator: "/")
+    return "\(baseUrl)/\(path)"
   }
 }
 
 public extension Endpoint {
+  static let locations = Self(baseUrl: cdnBaseUrl, pathComponents: ["locations"])
+  static let chatMessages = Self(baseUrl: cdnBaseUrl, pathComponents: ["messages"])
```
stephanlindauer, line 24: the CDN caches the response for 60s. i recommend using the api-gateway instead.

CriticalMapsKit/Sources/AppFeature/AppFeatureCore.swift
```diff
@@ -320,7 +372,17 @@ public struct AppFeature: ReducerProtocol {
...
         switch timerAction {
         case .timerTicked:
-          return EffectTask(value: .fetchData)
+          return .run { send in
+            await withThrowingTaskGroup(of: Void.self) { group in
+              group.addTask {
+                await send(.fetchLocations)
+              }
+              group.addTask {
+                await send(.fetchChatMessages)
```
stephanlindauer, line 381: just a question to understand how iOS is doing this: the desired outcome is that chat (post and get) and location (get) will only be run on demand, like only when the user opens that part of the app (map or chat window). the only thing that gets called on an interval (30sec) would be the locations put, is that right for that implementation as well?

CriticalMapsKit/Sources/ApiClient/Endpoint.swift
stephanlindauer: *just for messages though, locations should be cached via the CDN

mltbnz: ah. good point!. Didn't pay attention on that one

CriticalMapsKit/Sources/AppFeature/AppFeatureCore.swift
mltbnz: ok. the chat message get would be run on a timer tick (12 sec) as well as the getting the locations. The map is kind of open all the time due to the navigation structure we have but I could pause the timer when another view is openend of course. 
regarding the chat feature: I can update the behaviour to only fetch messages when the view is open but then we would loose the "unread messages" feature I guess since it wouldn't make a lot sense anymore. But I am not sure if anyone was into that one anyhow 😅

The location put is currently triggered on location updates. If a user moves more then 40 metres the client will send that request.

stephanlindauer: pausing once the map view is out of sight sounds like a good idea. there can be another call right after switching back to the map though.

stephanlindauer: why would the "unread messages" feature not work anymore? wouldn't that just be a matter of comparing the last shown messages to the ones that we added since then? alternatively there could be a timestamp that saves the last shown message. when there is new content everything with an older timestamp would be considered "new". does that make sense or am i missing the point?

mltbnz: since we are not getting message updates anymore when the view is not showing there will only be data available the user has already seen. the comparison behaviour from before is very close to what you suggested btw

CriticalMapsKit/Sources/AppFeature/RequestTimerCore.swift
```diff
@@ -2,7 +2,7 @@ import ComposableArchitecture
 import Foundation
 
 public struct RequestTimer: ReducerProtocol {
-  public init(timerInterval: Int = 12) {
+  public init(timerInterval: Int = 15) {
```
stephanlindauer, line 5:
```suggestion
  public init(timerInterval: Int = 30) {
```

stephanlindauer: we really need to lower the load on the API. as far as i understand this is the interval for sending the own location. can we set the interval to 30 seconds instead?

stephanlindauer (changes requested):

